PR #4, opened 2019-08-25
dwm: new package at 6.2

fanboimsft:
dynamic window manager from suckless.org

- [6.2] Latest upstream version.


community/dwm/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+# Amend install path to '/usr'
```
dylanaraps: Full stop `.`.

community/dwm/build
```diff
@@ -0,0 +1,10 @@
...
+sed -i'' "s#/usr/local#/usr#" config.mk
+
+# Remove pixel gap
```
dylanaraps: Full stop `.`.

community/dwm/build
```diff
@@ -0,0 +1,10 @@
...
+sed -i'' "s/static const int resizehints = 1/static const int resizehints = 0/" config.def.h
```
dylanaraps: Any specific reason for this change?

fanboimsft: Without the patch there is a weird 1pixel gap at the border of some windows. With the patch all windows just fill the screen as they should.

dylanaraps: :+1:

dylanaraps:
LGTM :+1:

PR #10, opened 2019-09-03
i3-gaps: 4.17.1

aidanharris:
## Description of package

The i3-gaps window manager:

screen

## Dependencies (excludes already packaged deps)

- [ ] i3status
- [x] [bsd-compat-headers](https://github.com/kisslinux/community/pull/11)
- [x] [libev](https://github.com/kisslinux/community/pull/12)
- [x] [libpcre](https://github.com/kisslinux/community/pull/13)
- [x] [libxkbcommon](https://github.com/kisslinux/community/pull/14)
- [x] [startup-notification](https://github.com/kisslinux/community/pull/15)
- [x] [xcb-util-cursor](https://github.com/kisslinux/community/pull/16)
- [x] [xcb-util-image](https://github.com/kisslinux/community/pull/17)
- [x] [xcb-util-renderutil](https://github.com/kisslinux/community/pull/18)
- [x] [xcb-util-xrm](https://github.com/kisslinux/community/pull/19)
- [x] [yajl](https://github.com/kisslinux/community/pull/20)

## Checklist


community/i3-gaps/build
```diff
@@ -0,0 +1,33 @@
...
+		then
+			continue
+		fi
+		patch -p1 < "${patch}"
+	done
+fi
+
+autoreconf -fvi
+
+./configure --prefix=/usr --enable-sanitizers
```
dylanaraps: Move each flag to a new line indented by 4 spaces.

Is `--enable-sanitizers` needed?

community/i3-gaps/build
```diff
@@ -0,0 +1,33 @@
...
+	do
+		if [ -z "${patch}" ] || [ ! -e "${patch}" ]
+		then
+			continue
+		fi
+		patch -p1 < "${patch}"
+	done
+fi
+
+autoreconf -fvi
```
dylanaraps: Is `autoreconf` necessary?

community/i3-gaps/build
```diff
@@ -0,0 +1,33 @@
...
+then
+	for patch in ${PATCHES}
+	do
+		if [ -z "${patch}" ] || [ ! -e "${patch}" ]
+		then
+			continue
+		fi
+		patch -p1 < "${patch}"
+	done
+fi
```
dylanaraps: Everything above this line can go. Patches should be done in a for loop if all `-p` match or one per line.

dylanaraps:
I think it's better if we just ship `i3-gaps`.

dylanaraps:
Also, any reason why `-std=gnu99` needs to be added to the `CFLAGS` in some of the PRs?

community/i3-gaps/build
aidanharris: Sanitizers shouldn't be needed but it could help find issues (Arch enables it by default). As a compromise I'll leave this unset by default but add an `$EXTRA_I3_CONF` variable (`./configure --prefix=/usr ${EXTRA_I3_CONF}` so users can pass extra options to configure if they want to (Gentoo does this for some packages and it basically allows you to do something like `env EXTRA_I3_CONF="--enable-sanitizers" kiss build i3-gaps` and then your variable gets expanded with any extra configuration options you wanted, you'd of course have to put that variable in a config file somewhere or wrapper script to make it persist).

community/i3-gaps/build
aidanharris: It's not strictly necessary but helps to make sure all of the auto-generated files can be built correctly and used without issues rather than relying on the (potentially old) `./configure` script that gets shipped with the tarball. Most distros run `autoreconf` on all projects that use autotools.

community/i3-gaps/build
dylanaraps: We don't currently add any package specific environment variables and I really don't want to start. The intended behavior for package customization is a user repository, however we're possibly exploring a `USE_FLAGS` kind of system for the future.

aidanharris:
@dylanaraps I used `-std=gnu99` because of some compile errors that were noted in Alpine's APKGBUILD's I was using as a reference. I'll do some more tests with `gcc` and `clang` and see if this is still necessary.

community/i3-gaps/build
dylanaraps: KISS doesn't run `autoreconf` unless the package doesn't provide a generated `configure` script. If `i3` provides `configure`, remove it. :+1:

aidanharris:
From a maintenance point of view the `i3-gaps` package is the same as the `i3-wm` package just with a different source so there's no real difficulty in having both (they do conflict though). I'll remove the `i3-wm` package, it can always be added back at a later date.

community/i3-gaps/build
```diff
@@ -6,7 +6,7 @@ export CHOST="${CHOST:-x86_64-pc-linux-musl}"
 
 patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
 
-./configure --prefix=/usr ${EXTRA_I3_CONF:-}
+./configure --prefix=/usr
```
dylanaraps: Split the flag onto a new line indented by 4 spaces.

community/i3-gaps/depends
```diff
@@ -0,0 +1,36 @@
...
+libXxf86vm
+libdrm
+libev
+libffi
+libpcre
+libpng
+libxcb
+libxkbcommon
+libxshmfence
+make
```
dylanaraps: This dependency is implied.

aidanharris:
@dylanaraps This should be okay now (unless I've missed something). I'll rebase it on top of master for a cleaner history without all of my fixup commits.

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+set -u
+
+export CHOST="${CHOST:-x86_64-pc-linux-musl}"
```
dylanaraps: KISS doesn't use `CHOST` and only supports `x86_64-pc-linux-musl`. Change the `cd` to go directly to `x86_64-pc-linux-musl`.

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
...
+
+patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
+
+./configure \
+    --prefix=/usr
+
+cd "${CHOST}" && make
```
dylanaraps: There's no need for `&&` as `-e` (`errexit`) is used. Move `make` on to its own line below.

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
...
+
+make DESTDIR="$1" PREFIX=/usr install
```
dylanaraps: Is `PREFIX` also needed during the install step?

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
...
+
+patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
+
+./configure \
+    --prefix=/usr
+
+cd "x86_64-pc-linux-musl"
+
+make
+
```
dylanaraps: Remove this blank line.

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+set -u
```
dylanaraps: Why is `set -u` needed?

community/i3-gaps/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
+
+./configure \
+    --prefix=/usr
+
+cd "x86_64-pc-linux-musl"
```
dylanaraps: This doesn't need quoting.

community/i3-gaps/build
```diff
@@ -2,6 +2,9 @@
 
 patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
 
+# This is needed for i3-gaps since they don't have release tarballs with pre-generated ./configure scripts
```
dylanaraps: This comment can be removed.

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
+
+# This is needed for i3-gaps since they don't have release tarballs with pre-generated ./configure scripts
+autoreconf -fvi
+
+./configure \
+    --prefix=/usr
```
dylanaraps: Add: 

```
--disable-docs
```

community/i3-gaps/depends
```diff
@@ -0,0 +1,35 @@
...
+startup-notification
+xcb-util
+xcb-util-cursor
+xcb-util-image
+xcb-util-keysyms
+xcb-util-renderutil
+xcb-util-wm
+xcb-util-xrm
+yajl
+zlib
```
dylanaraps: How was this dependency list created/generated?

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+patch -p1 < i3-4.16-musl-GLOB_TILDE.patch
+
+# This is needed for i3-gaps since they don't have release tarballs with pre-generated ./configure scripts
+autoreconf -fvi
```
dylanaraps: This should work with just `-i` no?

community/i3-gaps/build
```diff
@@ -0,0 +1,14 @@
...
+
+./configure \
+    --prefix=/usr
+
+cd x86_64-pc-linux-musl
```
dylanaraps: Remove this line and add `--disable-builddir` to `./configure`.

dylanaraps:
Any updates on this (and the other packages)?

dylanaraps:
Any update? Will close this (and the PRs tomorrow) otherwise.

dylanaraps:
Closed due to inactivity.

PR #11, opened 2019-09-03
bsd-compat-headers: 0.7.1

aidanharris:
## Description of package

These files are small enough it may be better to include it in `core` directly. This is what Alpine does (https://gitlab.alpinelinux.org/alpine/aports/tree/master/main/libc-dev)

## Checklist


dylanaraps:
Yeah, these are probably better as a part of the `musl` package. I'll make this change. :+1:

dylanaraps:
This can be closed, the `musl` package now includes the headers.

https://github.com/kisslinux/repo/commit/8fc7e1d1c537d85ded1275492f106ead9f5d3873

PR #12, opened 2019-09-04
libev: 4.16.1

aidanharris:
## Description of package

A high-performance event loop/event model with lots of features

## Checklist


community/libev/build
```diff
@@ -0,0 +1,34 @@
...
+then
+	for patch in ${PATCHES}
+	do
+		if [ -z "${patch}" ] || [ ! -e "${patch}" ]
+		then
+			continue
+		fi
+		patch -p1 < "${patch}"
+	done
+fi
```
dylanaraps: Everything above this (minus the `shebang` should be removed). Patches should be done with a for loop if `-p` matches for all of them or one per line if it doesn't.

community/libev/build
```diff
@@ -0,0 +1,34 @@
...
+
+sed -i -e "/^include_HEADERS/s/ event.h//" Makefile.am
+
+autoreconf -vfi
```
dylanaraps: Is `autoreconf` necessary (also the `sed` above it)?

community/libev/depends
```diff
@@ -0,0 +1 @@
+make
```
dylanaraps: We don't list `make` as a dependency, it's implied.

community/libev/build
```diff
@@ -0,0 +1,34 @@
...
+
+./configure \
+	--disable-maintainer-mode \
```
dylanaraps: 4 spaces for indentation.

community/libev/build
aidanharris: The `sed` is needed to prevent a conflict with libevent (both install `event.h`). I'll remove the `autoreconf`.

dylanaraps: Is this not a simpler solution?

```sh
# Goes after 'make install'.
rm -f "$1/usr/include/event.h"
```

community/libev/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+patch -p1 < libev-4.25-pc.patch
+
+sed -i -e "/^include_HEADERS/s/ event.h//" Makefile.am
+
+./configure \
+    --disable-maintainer-mode \
```
dylanaraps: What is the default value of `maintainer-mode`? If it's "disabled" or "off" this can be removed.

community/libev/build
```diff
@@ -0,0 +1,12 @@
...
+    --prefix=/usr
+
+make PREFIX=/usr
```
dylanaraps: Is `PREFIX` needed for `make`?

dylanaraps:
About this `libevent` conflict, what happens if `libevent` is **not** installed? Does `libev` still function with a missing header?

community/libev/build
aidanharris: Good catch:

> dnl maintainer-mode's default is 'disable' unless 'enable' is passed

This isn't needed.

dylanaraps:
The `depends` file can be removed if empty.

aidanharris:
@dylanaraps libev should still function without the header file. The header file itself contains the following comment:

> libevent compatibility header, only core events supported

Anything that needs this will probably just end up using libevent.

dylanaraps:
OK. We're safe to `rm` the conflicting header. :+1:

dylanaraps:
Note: The included patch likely does nothing at all without `autoreconf`. 

I'm creating an alternative method of generating the `pkg-config` file.

dylanaraps:
Amended build script. Removes the patch and the `event.h` header.

```sh
#!/bin/sh -e

./configure \
    --prefix=/usr

make
make DESTDIR="$1" install

mkdir -p "$1/usr/lib/pkgconfig"

# Grab package version.
read -r version _ < "${0%/*}/version"

# Create a 'pkg-config' file.
cat > "$1/usr/lib/pkgconfig/libev.pc" <<EOF
prefix=/usr
exec_prefix=\${prefix}
libdir=/usr/lib
includedir=\${prefix}/include

Name: libev
Description: A high-performance event loop/event model with lots of features
Version: $version
Libs: -L\${libdir} -lev
Libs.private:
Cflags: -I\${includedir}
EOF

# Remove compatibility header which conflicts with 'libevent'.
rm -f "$1/usr/include/event.h"
```

dylanaraps:
The creator of `libev` is against adding a `pkg-config` file so I don't think we need to add one either.

Source: http://lists.schmorp.de/pipermail/libev/2013q1/002128.html

In fact `i3` (and I imagine everything else) doesn't use `pkg-config` for `libev` so adding this file accomplishes nothing. 

`i3` `configure`:
 
```
# libev does not ship with a pkg-config file :(.
```

Lets remove it.

community/libev/checksums
```diff
@@ -0,0 +1,2 @@
+2d5526fc8da4f072dd5c73e18fbb1666f5ef8ed78b73bba12e195cfdd810344e  libev-4.27.tar.gz
+3dc1a3f74203c05c3e801ba54218eccb2e9bef73b7fc4400ba959aa82adf7a94  libev-4.25-pc.patch
```
dylanaraps: Checksums need updating.

dylanaraps:
For reference, Fedora also removed the `pkg-config` file:

```
    Don't use pkg-config to check for libev
    
    Several distributions used to ship a .pc file for libev, but upstream has
    ultimately rejected supporting pkg-config. This is causing distros to
    remove support for libev.
```

aidanharris:
@dylanaraps I'll remove it now. About the checksums, I did re-run `kiss c libev` but it didn't remove the patches from the checksums file. I'll do that manually now, (not sure if that's a bug in `kiss`).

dylanaraps:
>  (not sure if that's a bug in kiss).

I haven't seen this happen at all ever. Not sure why it didn't remove the unneeded checksum. The package manager loops over the `sources` file and when regenerating checksums it always overwrites the old ones.

dylanaraps:
LGTM now :+1:

dylanaraps:
The PR title doesn't seem to match the package version(?). Also is the source link related to upstream at all?

PR #14, opened 2019-09-04
xkbcommon: 0.8.4

aidanharris:
## Description of package

X.Org xkbcommon library

## Checklist


community/libxkbcommon/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+meson \
+	--prefix=/usr \
```
dylanaraps: 4 spaces for indentation.

community/libxkbcommon/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+meson \
+    --prefix=/usr \
+    -Denable-wayland=false \
+    -Denable-docs=false \
+    -Denable-x11=true \
```
dylanaraps: This is `true` by default.

community/libxkbcommon/build
```diff
@@ -0,0 +1,11 @@
...
+    -Ddefault_library=both build
```
dylanaraps: This isn't an option?

community/libxkbcommon/depends
```diff
@@ -0,0 +1,4 @@
+bison
+libXau
```
dylanaraps: This dependency doesn't exist in the build files from what I can see.

community/libxkbcommon/depends
```diff
@@ -0,0 +1,4 @@
...
+libxcb
+meson
```
dylanaraps: `meson make`

Also probably `bison make` (check with `ldd`).

PR #24, opened 2019-09-13
weechat: added gnutls

stefanfransen:
Sadly WeeChat requires GnuTLS for SSL/TLS connections.

This PR adds GnuTLS to WeeChat.

### DIFF
--- a/community/weechat/depends
+++ b/community/weechat/depends
@@ -1,5 +1,5 @@
 cmake     make
+gnutls
 libgcrypt
 libressl
 zlib
-gnutls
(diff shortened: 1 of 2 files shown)


community/weechat/depends
```diff
@@ -2,3 +2,4 @@ cmake     make
 libgcrypt
 libressl
 zlib
+gnutls
```
dylanaraps: Sort this alphabetically.

stefanfransen: Sorry about that. depends has been sorted.

Issue #35, opened 2019-10-01
ne out of date


dylanaraps:
Ping @sebastianertz 

> `3.1.2 -> 3.2.0`

PR #36, opened 2019-10-01
ne: bump to 3.2.0 #35

sebastianertz:
## Description of package

ne is a free text editor that runs on (hopefully almost) any UNIX-like machine. ne is easy to use for the beginner, but powerful and fully configurable for the wizard, and most sparing in its resource usage.

## Checklist


dylanaraps:
LGTM :+1:

PR #39, opened 2019-10-02
ne: bump to 3.2.1

sebastianertz:
## Checklist


dylanaraps:
Thanks :+1:

PR #40, opened 2019-10-03
mc: new package at 4.8.23

penguin-ff:
## Description of package

GNU Midnight Commander is a visual file manager with a feature rich full-screen text mode that allows you to copy, move and delete files and whole directory trees, search for files and run commands in the subshell. Internal viewer and editor are included

## Checklist


community/mc/sources
```diff
@@ -0,0 +1 @@
+http://ftp.midnight-commander.org/mc-4.8.23.tar.xz
```
dylanaraps: Is there an `https` source available?

penguin-ff: Found one - see my last commit

dylanaraps:
Is the new source an official mirror of `mc` or something third-party?

penguin-ff:
The former http url was i think the "official" one found on their website and also used by gentoo and arch. The new https url is "stolen" from the freebsd port. The checksums are the same (see commit)

dylanaraps:
If the `mc` developers don't provide a `https` mirror I'd much rather we stick to their official source even if it is `http`. I think this is an issue we should raise upstream, apologies for the run around.  :+1:

penguin-ff:
No problem. 
There is also a https link to github on their site, but this tarball has no configure script. You have the whole automake / autoconf mess and i think one need also autopoint from the gettext package (not packaged yet in kisslinux i think)
Is the pull request now ready to merge for you?

dylanaraps:
Swap it back to the original `http` link and I'll merge. :+1:

dylanaraps:
LGTM :+1:

Issue #41, opened 2019-10-05
nano out of date


dylanaraps:
Ping @sebastianertz 

`4.4 -> 4.5`

PR #42, opened 2019-10-05
nano: bump to 4.5 #41

sebastianertz:
## Checklist


dylanaraps:
LGTM :+1:

PR #46, opened 2019-10-09
new package: libwm

nikotidar:
## Checklist


community/libwm/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+
+sed -i'' "s#/usr/local#/usr#" config.mk
+
+make
+make DESTDIR="$1" install
```
dylanaraps: This can just be:

```sh
make

install -D libwm.a "$1/usr/lib/libwm.a"
install -D wm.h    "$1/usr/include/wm.h"
```

We can then avoid patching the Makefile.

konimex: or just `make DESTDIR="$1" PREFIX=/usr install` should be good too.

dylanaraps:
Ping @nikotidar

nikotidar:
> Ping @nikotidar

Hi, Dylan,

Sorry for late reply, my second laptop is broken. I'll be back 4 days later.
Thanks for your help!

dylanaraps:
All good. :)

PR #47, opened 2019-10-12
emacs: new package at 26.3

penguin-ff:
## Description of package

GNU Emacs

## Checklist


dylanaraps:
LGTM :+1:

I'll land the package manager update first and will merge this afterwards.

community/emacs/sources
```diff
@@ -0,0 +1 @@
+https://gnuftp.uib.no/emacs/emacs-26.3.tar.xz
```
dithpri: Shouldn't this be either the [main GNU ftp](https://ftp.gnu.org/gnu/emacs/emacs-26.3.tar.xz) or an auto-redirect using http://ftpmirror.gnu.org/emacs/?
I don't think it makes sense to use one mirror for every KISS download.

penguin-ff: You're right. I'll take "your" first link...

dylanaraps:
Thanks :+1:

PR #52, opened 2019-10-16
zsh: new package at 5.7.1

dithpri:
## Checklist


community/zsh/patches/zsh_tests.patch
```diff
@@ -0,0 +1,72 @@
+--- a/Test/A01grammar.ztst
++++ b/Test/A01grammar.ztst
+@@ -95,13 +95,14 @@
+ 0:`exec' with -l option
+ >-zsh
+ 
+-  (exec -a /bin/SPLATTER /bin/sh -c 'echo $0')
+-0:`exec' with -a option
+->/bin/SPLATTER
++# The following tests don't work if using busybox (symlinks)
```
dylanaraps: Remove this patch.

community/zsh/build
```diff
@@ -0,0 +1,29 @@
...
+    --enable-runhelpdir=/usr/share/zsh/help \
+    --enable-scriptdir=/usr/share/zsh/scripts \
+    --enable-additional-fpath=/usr/share/zsh/vendor-completions \
+    --enable-libc-musl \
+    --disable-gdbm \
+    --with-term-lib='ncursesw' \
+    --enable-cap
+
+make
+make test
```
dylanaraps: Remove `make test`

community/zsh/build
```diff
@@ -0,0 +1,29 @@
...
+make DESTDIR="$1" install
```
dylanaraps: Add a new line below this.

dithpri:
Done.

community/zsh/build
```diff
@@ -0,0 +1,28 @@
...
+./configure \
+    --prefix=/usr \
+    --enable-function-subdirs \
+    --enable-fndir=/usr/share/zsh/functions \
+    --enable-runhelpdir=/usr/share/zsh/help \
+    --enable-scriptdir=/usr/share/zsh/scripts \
+    --enable-additional-fpath=/usr/share/zsh/vendor-completions \
+    --enable-libc-musl \
+    --disable-gdbm \
+    --with-term-lib='ncursesw' \
```
dylanaraps: Quotes here aren't needed.

community/zsh/sources
```diff
@@ -0,0 +1,3 @@
+https://sourceforge.net/projects/zsh/files/zsh/5.7.1/zsh-5.7.1.tar.xz
+patches/remove_bin_duplicate.patch
+files/zprofile
```
dylanaraps: Is shipping a `zprofile` file a typical part of packaging `zsh`?

community/zsh/depends
```diff
@@ -0,0 +1,3 @@
+make make
```
dylanaraps: `make` is an implicit dependencies and doesn't have to be listed.

community/zsh/depends
```diff
@@ -0,0 +1,3 @@
...
+xz make
```
dylanaraps: `xz` should go after `ncurses` (alphabetical ordering is desired though it sometimes makes sense to order things differently).

community/zsh/sources
```diff
@@ -0,0 +1,3 @@
+https://sourceforge.net/projects/zsh/files/zsh/5.7.1/zsh-5.7.1.tar.xz
+patches/remove_bin_duplicate.patch
```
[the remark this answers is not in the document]
dithpri: Both `zsh` and `zsh-5.7.1`, hardlinked. Seems this is used to perform a safe upgrade when zsh is installed directly from source without packaging.

community/zsh/sources
dithpri: It seems so. I know Gentoo, Arch and even Alpine provide a zprofile. The one I included will just include `/etc/profile`.

community/zsh/files/zprofile
```diff
@@ -0,0 +1 @@
+emulate sh -c 'source /etc/profile'
```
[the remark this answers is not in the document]
dithpri: Yup, I guess that's why it works. Didn't realize `source` was a bash-ism, thanks! Will change.

dithpri:
If you have no further recommendations/objections, I'm ready to squash these commits before merging.

community/zsh/build
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh -e
+
+patch -p1 < remove_bin_duplicate.patch
+
+for _fpath in AIX BSD Cygwin Darwin Debian Mandriva openSUSE Redhat Solaris; do
+    rm -rf Completion/$_fpath
```
dylanaraps: Quote the argument with the variable.

community/zsh/build
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh -e
+
+patch -p1 < remove_bin_duplicate.patch
+
+for _fpath in AIX BSD Cygwin Darwin Debian Mandriva openSUSE Redhat Solaris; do
```
dylanaraps: Rename this to `path`.

community/zsh/sources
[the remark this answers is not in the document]
dylanaraps: I think it's better we include it. We already do so for any other package which provides a `-version` binary. There may also be a symlink configure option though I'm unsure.

community/zsh/build
dithpri: That's not necessary in this case, but sure.

community/zsh/build
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh -e
+
+printf "functions='Completion/*comp* Completion/Base/*/* Completion/Linux/*/* Completion/Unix/*/* Completion/X/*/* Completion/Zsh/*/*'\n" >> Src/Zle/complete.mdd
```
dylanaraps: Something like this is also an option:

```sh
{
    printf "functions='"
    printf "Completion/*comp* "
    printf "Completion/Base/*/* "
    printf "Completion/Zsh/*/* "
    printf "Completion/Linux/*/* "
    printf "Completion/Unix/*/* "
    printf "Completion/X/*/* "
    printf "'\\n"
} >> Src/Zle/complete.mdd
```

This also makes adding/removing entries a little easier. Though a loop may turn out even simpler. ;)

dylanaraps: Another:

```sh
    printf '%s ' "functions='" "Completion/*comp*"    "Completion/Base/*/*" "Completion/Zsh/*/*" \
                               "Completion/Linux/*/*" "Completion/Unix/*/*" "Completion/X/*/*" "'
" >> Src/Zle/complete.mdd
```

dylanaraps: Or even:

```sh
    printf '%s ' "functions='" \
                 "Completion/*comp*"    \
                 "Completion/Base/*/*" \
                 "Completion/Zsh/*/*" \
                 "Completion/Linux/*/*" \
                 "Completion/Unix/*/*" \
                 "Completion/X/*/*" "'" >> Src/Zle/complete.mdd
```

dylanaraps: A loop: 

```sh
printf "functions='Completion/*comp*'\\n" >> Src/Zle/complete.mdd

for comp in Base Zsh Linux Unix X; do
    printf "functions+=' Completion/$comp/*/*'\\n" 
done >> Src/Zle/complete.mdd
```

(Untested)

dylanaraps: Another loop:

```sh
{
    printf "functions='Completion/*comp* " 

    for comp in Base Zsh Linux Unix X; do
        printf "Completion/$comp/*/* " 
    done >> Src/Zle/complete.mdd

    printf "'"
} >> Src/Zle/complete.mdd
```

PR #68, opened 2019-11-07
Added suckless slock

p410n3:
## Simple X display locker. 

This is the simplest X screen locker we are aware of. It is stable and quite a lot of people in our community are using it every day when they are out with friends or fetching some food from the local pub.

## Notes from me

This is my first package, I hope all is fine with it :)

## Checklist

- [ x ] Latest upstream version.
- [ x ] I agree to maintain this package (**required**).
- [ x ] I agree to notify KISS if I can no longer maintain this package (**required**).


community/slock/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+sed 's/usr\/local/usr/g' config.mk
```
dylanaraps: I take it you want to edit the file in-place? You can also use any character as a separator in `sed`. Saves having to escape the forward-slash.

```sh
sed -i'' 's|usr/local|usr|g' config.mk
```

dylanaraps:
LGTM :+1:

p410n3:
Yeah I forgot the -i flag too ... oops

PR #69, opened 2019-11-07
Added suckless Farbfeld

p410n3:
## Farbfeld
farbfeld is a lossless image format which is easy to parse, pipe and compress. It is also a dependencie to suckless sent
## Checklist

- [ x ] Latest upstream version.
- [ x ] I agree to maintain this package (**required**).
- [ x ] I agree to notify KISS if I can no longer maintain this package (**required**).


p410n3:
Shit it added both commits to one PR. lemme redo that

PR #73, opened 2019-11-13
crystal : new package at 0.31.1

OgataKaede:
## Description of package

This binary version of Crystal Language compiler and it utilities, i had no idea to build since this packages is self-hosted in it language. need crystal to build crystal :) 

## Checklist


konimex:
The problem of `crystal` here is not that far off from the `rust` problem. However, we can't just trust binaries outright. So a bootstrap *will* be needed if `crystal` isn't already installed.

I suggest you to take a look at how Void or Alpine (since they're, in a sense, the closest ones to KISS) Linux builds `crystal`.

PR #76, opened 2019-11-14
spotifyd: new package at 0.2.20

alaughlin:
## Description of package
Spotifyd allows third party Spotify clients to be able to stream music without relying on the official client

## Checklist


community/spotifyd/build
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+cargo build --release --no-default-features --features alsa_backend
```
dylanaraps:
```sh
 cargo build \
    --release \
    --no-default-features \
    --features alsa_backend
```

community/spotifyd/depends
```diff
@@ -0,0 +1,2 @@
+rust make
```
dylanaraps:
```sh
alsa-lib make
rust      make
```

community/spotifyd/build
alaughlin: fixed

community/spotifyd/depends
alaughlin: fixed

dylanaraps:
Thanks. :+1:

PR #77, opened 2019-11-15
cmus: new package at 2.8.0

alaughlin:
## Description of package
cmus is a terminal-based audio player
## Checklist


dylanaraps:
LGTM :+1:

PR #78, opened 2019-11-16
libmad: new package at 0.15.1b

alaughlin:
## Description of package
mpeg decoder used by audio applications such as cmus

## Checklist


community/libmad/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+for patch in *.patch; do
```
dylanaraps: Change this entire block to:

```sh
patch -p1 < remove-fforce-mem.patch
```

community/libmad/build
```diff
@@ -0,0 +1,11 @@
...
+    patch -p1 < "$patch"
+done
+
+./configure \
+    prefix=/usr
```
dylanaraps: `--prefix=/usr`?

alaughlin: either seems to work. but I'll change it :)

community/libmad/build
dylanaraps: Actually, you can remove the patch in its entirety and use:

```sh
sed -i'' '/ -fforce-mem/d' configure
```

community/libmad/build
alaughlin: fixed :)

community/libmad/build
alaughlin: good call! fixed :)

dylanaraps:
Looks good. :+1:

PR #80, opened 2019-11-16
yajl: new package at 2.1.0

alaughlin:
## Description of package
https://lloyd.github.io/yajl/

## Checklist


community/libyajl/build
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+./configure \
+    -p /usr
```
alaughlin: `--prefix` doesn't work for this package

dylanaraps:
I would name this package `yajl`.

Also, `--prefix` looks to be supported: https://github.com/lloyd/yajl/blob/master/configure#L26

alaughlin:
I can rename it.

I've tried `--prefix=/usr` and `--prefix /usr` and neither seemed to work 🤷‍♂

dylanaraps:
OK, `-p` is fine then. :+1:

alaughlin:
@dylanaraps renamed to `yajl` 👍

PR #81, opened 2019-11-16
libxkbcommon: new package at 0.9.1

alaughlin:
## Description of package
https://github.com/xkbcommon/libxkbcommon

## Checklist


community/libxkbcommon/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+mkdir build
```
dylanaraps: These two lines can be removed.

This is what the `atk` package looks like for reference:

```sh
export DESTDIR="$1"

meson \
    --prefix=/usr \
    --sysconfdir=/etc \
    --mandir=/usr/share/man \
    -Ddocs=false \
    -Dintrospection=false \
    . output

ninja -C output
ninja -C output install
```

community/libxkbcommon/depends
```diff
@@ -0,0 +1,2 @@
+meson   make
+samurai make
```
dylanaraps: This can be removed. `meson` pulls in `samurai` automatically.

alaughlin: 👍

community/libxkbcommon/build
alaughlin: fixed!

community/libxkbcommon/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+export DESTDIR="$1"
+
+meson \
+    --prefix=/usr \
+    --sysconfdir=/usr \
```
dylanaraps: This and `--mandir` may not need adding if these values match the defaults.

alaughlin: you're right. fixed!

Issue #67, opened 2019-11-06
Add xorriso to repo


scsijon:
[quote]xorriso copies file objects from POSIX compliant filesystems into Rock Ridge enhanced ISO 9660 filesystems and allows session-wise manipulation of such filesystems. It can load the management information of existing ISO images and it writes the session results to optical media or to filesystem objects.
Vice versa xorriso is able to copy file objects out of ISO 9660 filesystems. [/quote]

I couldn't say it any better, so I didn't try.

url is https://www.gnu.org/software/xorriso/

Talk to Thomas Schmitt,  scdbackup@gmx.net. I'm sure he would help and support it and help you in adding it as he has with many others.

Oh, and it is both command line and gui capable.

dylanaraps:
Transferred. :+1:

dylanaraps:
Added thanks to #84.

scsijon:
On 17/11/19 07:25, dylan wrote:
> Closed #67.
> 
Just a note that xorriso is also known as libisoburn (which contains it) 
so I don't know which name it should really be under. ?Do we need the 
full package for xorriso or the libisoburn version? Yes confusing I 
know, it's just how Thomas Schmitt did it back then, and both names are 
relevant. Maybe a note somewhere?

regards

dylanaraps:
Repology lists it as `xorriso` as it seems to be named this way in all major distributions.

scsijon:
On 17/11/19 11:05, dylan wrote:
> Repology lists it as `xorriso` as it seems to be named this way in all major distributions.
> 
  No worries, I was just mentioning what Thomas had said in his email to 
us in T2 a few weeks ago when it was queeried by me. I'm cc'ing him in 
case he wants to reply to this list also.

To quote him
-----------------
My apologies for creating this ambiguity. The reason lies back in 2008,
when the first packager preferred the 2-in-1 source tarball, with the
reasoning that this worked well with cdrskin in the libburn tarball.
Because of the difficulty to install .so without breaking the local
package management, i always had a xorriso-standalone tarball, which in
2010 became GNU xorriso.

Distros which use both, libisoburn and xorriso tarballs, most probably
have a duplication. Who installs libisoburn should also install its
xorriso binary. The size difference is substantial.

But as soon as the user needs a newer xorriso than provided by the 
distro, it is time for GNU xorriso from source. It does not need 
installation to places like /usr/bin, but can be run at the place were 
it was built.
[...]

PR #84, opened 2019-11-16
xorriso: new package at 1.5.2

alaughlin:
## Description of package
https://www.gnu.org/software/xorriso/#download

https://github.com/kisslinux/community/issues/67

## Checklist


community/xorriso/depends
```diff
@@ -0,0 +1,3 @@
+bzip2
+musl
```
dylanaraps: Musl as a dependency doesn't need to be added as it will always exist. :)

alaughlin: duh! thank you. fixed :)

PR #85, opened 2019-11-17
sdl2: new package at 2.0.10

alaughlin:
## Description of package
https://www.libsdl.org/index.php

## Checklist


community/sdl2/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+patch -p1 < SDL2-2.0.10-opengl_include_fix-1.patch
```
alaughlin: note: LFS seems to think this is a required patch for i686 systems: http://www.linuxfromscratch.org/blfs/view/svn/multimedia/sdl2.html

if you don't think we need it, I'll remove it

dylanaraps: We don't support i686 systems, remove it. :+1:

alaughlin: removed!

dylanaraps:
Thanks!

PR #86, opened 2019-11-17
sdl: new package at 1.2.15

alaughlin:
## Description of package
https://www.libsdl.org/download-1.2.php

## Checklist


community/sdl/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+sed -e '/_XData32/s:register long:register _Xconst long:' \
```
dylanaraps: What does this do?

alaughlin: fixes a bug that prevents it from compiling with x11 support, I added a comment to the build file

here's some relevant info:
https://www.linuxquestions.org/questions/linux-from-scratch-13/can%27t-compile-sdl-4175469154/

The above thread includes a link to a patch, but LFS has the simpler solution I ended up using:
http://www.linuxfromscratch.org/blfs/view/svn/multimedia/sdl.html

I figured a simple `sed` was better than a patch file?

dylanaraps: Sounds good. :+1:

Was just wondering. :)

alaughlin: :)

community/sdl/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+# fixes bug that prevents SDL from building with X11 support
```
dylanaraps:
```
# Fixes bug that prevents SDL from building with X11 support.
```

alaughlin: fixed 😆

dylanaraps:
Thanks :)

PR #93, opened 2019-12-01
Add qemu

jdavies-dev:
## Description of package
QEMU  emulation software

## Checklist

- [ Yes (4.1.1) ] Latest upstream version.
- [ Yes ] I agree to maintain this package (**required**).
- [ Yes ] I agree to notify KISS if I can no longer maintain this package (**required**).


jdavies-dev:
Tested on ppc64le only.

community/qemu/depends
```diff
@@ -0,0 +1,3 @@
+ncurses
+sdl
+make
```
dylanaraps: `make` can be removed. It's a given that it be installed.

community/qemu/build
```diff
@@ -0,0 +1,29 @@
...
+patch -p0 < 002-ppc.patch
+patch -p0 < 003-sigtrm-max.patch
+patch -p0 < 004-syscall.patch
+patch -p0 < 005-elfload.patch
+
+./configure \
+    --prefix=/usr \
+    --localstatedir=/var \
+    --sysconfdir=/etc \
+    --python=/usr/bin/python3 \
```
dylanaraps: Is `python` needed for build? If so, it should be a dependency.

community/qemu/build
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh -e
+
+patch -p0 < 001-map-sync.patch
+patch -p0 < 002-ppc.patch
+patch -p0 < 003-sigtrm-max.patch
+patch -p0 < 004-syscall.patch
+patch -p0 < 005-elfload.patch
```
dylanaraps:
```sh
for patch in *.patch; do
    patch -p0 < "$patch"
done
```

community/qemu/build
```diff
@@ -0,0 +1,29 @@
...
+    --python=/usr/bin/python3 \
+    --disable-glusterfs \
+    --disable-debug-info \
+    --disable-bsd-user \
+    --disable-werror \
+    --disable-xen \
+    --enable-kvm \
+    --enable-lzo \
+    --enable-sdl \
+    --enable-iconv \
```
dylanaraps: Is this vendored in or is it a needed dependency?

community/qemu/build
```diff
@@ -0,0 +1,29 @@
...
+    --localstatedir=/var \
+    --sysconfdir=/etc \
+    --python=/usr/bin/python3 \
+    --disable-glusterfs \
+    --disable-debug-info \
+    --disable-bsd-user \
+    --disable-werror \
+    --disable-xen \
+    --enable-kvm \
+    --enable-lzo \
```
dylanaraps: Is this vendored in or is it a needed dependency? (Assuming this is the compression lib)

community/qemu/depends
jdavies-dev: Make now removed.

community/qemu/build
jdavies-dev: Added python as a dependency.

community/qemu/build
jdavies-dev: Done

dylanaraps:
LGTM :+1:

Is this ready to merge?

community/qemu/build
jdavies-dev: I don't think lzo and sdl are vendored in.  Build fails if sdl2 or lzo are not installed and these options are used, and I don't think the qemu  source includes its own copies of these libs.

However, I removed "--enable-lzo" from the config, since ./configure will include support for lzo if the package is installed, and not if not.    The only ones I wanted to enable explicitly were sdl and ncurses, so I didn't specify for the others.  So that's why I added ncurses and sdl to depends.

[Note, for checking if a lib is vendored in, I am checking by installing/removing packages, the running builds, and just grepping through the source.  Do you have a better way?]

jdavies-dev:
Sure, if you're happy with it.  I have no further changes queued here.

community/qemu/build
dylanaraps: > [Note, for checking if a lib is vendored in, I am checking by installing/removing packages, the running builds, and just grepping through the source. Do you have a better way?]

I was just curious. You typically know post-build if the resulting library or binaries link to the dependencies or not (check with `ldd /path/to/file`). 

So if `cmus` was built with `ncurses`  and `ncurses` is then missing from `ldd /usr/bin/cmus`, it's most likely statically compiled into `cmus` and hence vendored in (cmus as an example). 

There are other ways of checking too. The dependency checker post build will not include anything vendored in as it parses `ldd` (so long as the dependency wasn't manually added to the depends file prior). 

Build output or its documentation are other methods of checking. You'll know for instance if the build starts downloading from the net.

Finally, as a last resort, there's the source code of the software.

dylanaraps:
Thanks :+1:

PR #94, opened 2019-12-03
xbacklight: new package at 1.2.3

lieuxnoir:
## Checklist


dylanaraps:
:+1:

PR #109, opened 2019-12-19
new package: mksh R57

a-schaefers:
## Description of package
An excellent shell in the spirit of KISS 
## Checklist


community/mksh/build
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+CC="gcc -static" sh Build.sh
+
+mkdir -p "$1/usr/bin"
```
dylanaraps: Remove this line.

community/mksh/build
```diff
@@ -0,0 +1,7 @@
...
+
+install -m755 "mksh" "$1/usr/bin/mksh"
```
dylanaraps:
```sh
 install -Dm755 "mksh" "$1/usr/bin/mksh"
```

This will also create the directory.

community/mksh/depends
```diff
@@ -0,0 +1 @@
+make
```
dylanaraps: `make` is an implicit dependency and doesn't have to be listed.

a-schaefers:
Oh boy, did this one quite a bit. I'll take care of it. When you see a redundant issue like this, I'll try to remember to apply it across everything, that way in your review you don't have to keep saying the same thing over and over. Thanks

a-schaefers:
fixed.

dylanaraps:
The depends file can be removed. :+1:

a-schaefers:
> The depends file can be removed

done 779c5c1

PR #112, opened 2019-12-19
new package: coreutils 8.31

a-schaefers:
## Checklist


a-schaefers:
gnucoreutils renamed to just "coreutils" and everything is in /usr/bin prefixed with 'g'

done

community/coreutils/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+export CC="gcc -static"
+
+export FORCE_UNSAFE_CONFIGURE=1
+
+./configure \
+    --prefix=/opt/gnu/coreutils \
+    --program-prefix=g \
+    --program-transform-name=s/stat/nustat/
```
dylanaraps: `nustat` -> `gnustat`? Or even `gstat`(?). I'm not sure how the name of this is typically handled when it conflicts. The BSDs might be a good place to look.

community/coreutils/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+export CC="gcc -static"
+
+export FORCE_UNSAFE_CONFIGURE=1
```
dylanaraps: Why is this needed?

community/coreutils/build
a-schaefers: > `nustat` -> `gnustat`? Or even `gstat`(?). I'm not sure how the name of this is typically handled when it conflicts. The BSDs might be a good place to look.

I  copy and pasted it from the freebsd make.conf actually, I'm sure they have a good reason for it.

community/coreutils/build
a-schaefers: It is my opinion static linking with musl is the way to go as much as possible. Is this not the KISS position?

For example stal.li linux has some short write-ups on the subject re: the benefits of static...

> Executing statically linked executables is much faster, because there are no expensive shared object lookups during exec().
> Statically linked executables are portable, long lasting and fail safe to ABI changes -- they will run on the same architecture even in 10 years time. Never expect errors like /lib/ssa/libstdc++.so.6:version 'GLIBCXX_3.4.4' not found again.
> Statically linked executables use less disk space. Most executables use only a small subset of the functions provided by a static library -- so there is absolutely no reason to link complete static libraries into a static executable (e.g. spoken for a hello_world.c you only need to link vprintf statically into the executable, not the whole static libc!). The contrary is true for dynamic libraries -- you always use the whole library, regardless what functions you are using.
> Statically linked executables consume less memory because their binary size is smaller and they only map the functions they depend on into memory (contrary to dynamic libs).
http://web.archive.org/web/20090525150626/http://blog.garbe.us/2008/02/08/01_Static_linking/

Also see https://sta.li/faq/

a-schaefers: @dylanaraps or does your question refer to the line:
`export FORCE_UNSAFE_CONFIGURE=1`
?

a-schaefers: export FORCE_UNSAFE_CONFIGURE=1

is required by configure because we are building as the root user in KISS and gnu does not feel that doing that is a good idea.

dylanaraps: In KISS the package manager should be run as your normal user actually. Only the install/remove package manager operations need root. 

Yeah, I was referring to `FORCE_UNSAFE_CONFIGURE=1`.

a-schaefers: Hmmm. I much prefer running all of kiss commands as root and not typing my su password all the time ... :( Can we not keep this environment variable?

dylanaraps: If you install `sudo` it will only prompt once for the password as `sudo` saves your auth for a period of time by default. I'm fine with keeping it around for your use case though.

community/coreutils/build
```diff
@@ -0,0 +1,12 @@
...
+
+./configure \
+    --prefix=/opt/gnu/coreutils \
```
dylanaraps: Can we not install to `/usr/bin` now as all programs are prefixed with `g`?

community/coreutils/build
a-schaefers: thanks, please keep it around. I'm anti-sudo. lol

community/coreutils/build
dylanaraps: Sounds good to me then. :+1:

dylanaraps:
The depends file can be removed too.

community/coreutils/build
a-schaefers: ERRG ARGH ACK Okay  got it!!!

a-schaefers:
should be all 'g'ood now

PR #113, opened 2019-12-19
new package: nawk 20180827

a-schaefers:
## Description of package
one true awk
## Checklist


dylanaraps:
Is this package really needed? I think it's better suited for a user repository which also includes a modified busybox package to exclude busybox awk so it can be installed to `/usr/bin`.

a-schaefers:
awk is a classic and this is the original, preferred by a lot of old school devs and bsd guys. totally different than gawk. Should definitely be in our repo, especially considering how easy it is to compile and literally no maintenance at all to upkeep the build.

a-schaefers:
We  don't want to modify the busybox base for the root user unless it is done for everybody, because too many variables come into play with maintaining packages / debugging failed builds, and also people's systems may break

konimex:
FWIW the `awk` in many systems is packaged as `nawk`. See [Repology](https://repology.org/project/nawk/versions).

a-schaefers:
nawk is fine by me

dylanaraps:
Happy with `nawk` :+1:

a-schaefers:
done

dylanaraps:
The depends file can be removed. :+1:

a-schaefers:
d584259 done

PR #114, opened 2019-12-19
new package: gawk 5.0.1

a-schaefers:
## Checklist


dylanaraps:
I'd rather we didn't use `/opt` at all. Can this not simply be installed as `gawk`?

a-schaefers:
I'm open to other ideas, but where do you propose we put all this stuff, then without breaking busybox?

I have packaged not only gawk, but also awk, and gnucoreutils, findutils, diffutils, and sbase and ubase. All of which will conflict busybox if installed to to /usr/bin

I'm open to other locations, but let's just make sure it's consistently implemented.

a-schaefers:
And yes we could simply install this as /usr/bin/gawk , let me know if you'd prefer that, but however, the rest of the stuff still needs to be somewhere either 1. later on PATH than /bin and /usr/bin . or 2. off of PATH completely (my preference is definitely the latter)

a-schaefers:
one way to do this if you prefer no /opt would be to install all of these tools with a 'g' prefix
https://www.freshports.org/sysutils/coreutils/
that's the approach freebsd takes

However, the down side is losing the ability to quickly modify PATH for a user. Currently, for example my PATH for my user account "adam" looks like this:

/home/adam/.nix-profile/bin:/opt/gnu/coreutils/bin:/opt/gnu/findutils/bin:/opt/gnu/diffuti
ls/bin:/opt/gnu/gawk/bin:/opt/gnu/grep/bin:/opt/gnu/patch/bin:/opt/awk/bin:/home/adam/bin/bin:/usr/bin:/bin:/usr/bin

if i did this to root I'd have a broken system, but it's fine for the regular user, and it gives me a better Emacs experience, because a lot of Emacs packages expect gnu tools, sadly. But i'm pretty locked in to Emacs, and busy box with Emacs is a nightmare.

konimex:
>but then the user would have to symlink them all manually, one by one, into his own specially directory full of symlinks, assuming the user wanted to do what I am currently doing. A pain, but not the end of the world at all.

At that point I think it's better to use user repository for that. The main package will use the `g` prefix to avoid conflicts with packages from main repo, then you'd use symlinks in your personal pkgdir (e.g. `/path/to/repo/coreutils/versions -> /var/db/kiss/community/versions`). So you'd only have to maintain your `build` file in your personal repo, and maintain the "upstream" `build`, `version`, `checksums`, `depends` (if any), and `sources` in the `community` repo.

a-schaefers:
Okay, sounds like the 'g' prefix may be the goto solution for this. @dylanaraps what say you? If so, I'll get to work on it, and I'll also work on getting the rest of the suckless and gnu software out of /opt (so need to bring it up again, if this is how we choose to resolve this one.)

dylanaraps:
Yup. I agree with the `g` prefix.

a-schaefers:
fixed

dylanaraps:
The depends file can be removed now. :)

a-schaefers:
done 23e4c2c

PR #115, opened 2019-12-19
new package: findutils 4.7.0

a-schaefers:
## Checklist


a-schaefers:
0ba33dd should be good to go

PR #117, opened 2019-12-19
new package: gtar 1.32

a-schaefers:
## Checklist


a-schaefers:
this should be ready

a-schaefers:
rename gnutar ==> gtar

all of our gnu utilities are now following the same conventions FreeBSD uses

community/gtar/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+export FORCE_UNSAFE_CONFIGURE=1
+export CC="gcc -static"
+
+./configure \
+    --prefix=/usr \
+    --program-prefix=g
+
+make DESTDIR="$1" install
```
dylanaraps: Missing standalone `make`?

a-schaefers: done

PR #118, opened 2019-12-19
new package: diffutils 3.7

a-schaefers:
## Checklist


a-schaefers:
should be good to go here as well now

community/diffutils/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+export CC="gcc -static"
+
+./configure \
+    --prefix=/usr \
+    --program-prefix=g
+
+make DESTDIR="$1" install
```
dylanaraps: Missing standalone `make`?

PR #119, opened 2019-12-19
new package: patch 2.7

a-schaefers:
## Checklist


a-schaefers:
fixed

community/patch/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+export CC="gcc -static"
+
+./configure \
+    --prefix=/usr \
+    --program-prefix=g
+
+make DESTDIR="$1" install
```
dylanaraps: Missing standalone `make`?

a-schaefers: i did not know we preferred that. will do

a-schaefers: I did notice once in a blue moon a package will fail without the standalone make. *cough* iptables

a-schaefers: done

PR #125, opened 2019-12-22
New Package: imagemagick6 6.9.10-80

a-schaefers:
## Checklist


community/libmagick6/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+./configure \
+    --prefix=/usr \
+    --sysconfdir=/etc \
+    --with-modules
+
+sed -i -e 's/ -shared / -Wl,-O1,--as-needed\0/g' libtool
```
dylanaraps: Reason for this change?

community/libmagick6/build
```diff
@@ -0,0 +1,10 @@
...
+
+make DESTDIR="$1" install
```
dylanaraps: Missing standalone `make`?

community/libmagick6/depends
```diff
@@ -0,0 +1,8 @@
+make
```
dylanaraps: `make` can be removed.

community/libmagick6/depends
```diff
@@ -0,0 +1,8 @@
...
+fontconfig
+libtool
```
dylanaraps: `libtool make`.

dylanaraps:
Should this package not be called `imagemagick`?

community/libmagick6/build
a-schaefers: monkey read arch pkgbuild monkey do. 

imagemagick was complex to build. 

maybe we don't need this section after all?

a-schaefers: How about we remove and see how long it lasts before we need it

a-schaefers:
now look at things

community/libmagick6/build
dylanaraps: My `imagemagick7` package in my personal repository is a simple `./configure; make; make install`. Nothing special needed. I say remove it.

Another question, why not ImageMagick7 (the latest version)?

a-schaefers: I need imagemagick6 for Emacs, and it is still maintained upstream so it's fine

PR #126, opened 2019-12-22
new package: libconfig 1.7.2

a-schaefers:
## Checklist


community/libconfig/sources
```diff
@@ -0,0 +1 @@
+https://github.com/hyperrealm/libconfig/archive/v1.7.2.tar.gz
```
dylanaraps: This source should remove the need for `autoreconf`.

```
https://hyperrealm.github.io/libconfig/dist/libconfig-1.7.2.tar.gz
```

community/libconfig/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+export MAKEINFO=true
+
+autoreconf
```
dylanaraps: With the source change these should be removable.

community/libconfig/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+export MAKEINFO=true
```
dylanaraps: Why is this needed again?

a-schaefers: lol i thought i had fixed all these

a-schaefers:
ready

PR #127, opened 2019-12-22
New package: uthash 2.1.0

a-schaefers:
## Checklist


community/uthash/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+mkdir -p "$1/usr/include"
+
+install -m755 src/* "$1/usr/include"
```
dylanaraps: I'm not sure if these should be `644` or `755` but this is an alternative way of installing it:

```sh
for file in src/*.h; do
    install -Dm644 "$file" "$1/usr/include/${file##*/}"
done
```

PR #130, opened 2019-12-22
new package: sbase

a-schaefers:
## Checklist


community/sbase/depends
```diff
@@ -0,0 +1 @@
+make
```
dylanaraps: This file can be removed. :+1:

a-schaefers:
this is ready

PR #131, opened 2019-12-22
new package: ubase

a-schaefers:
## Checklist


community/ubase/depends
```diff
@@ -0,0 +1 @@
+make
```
dylanaraps: This file can be removed.

community/ubase/version
```diff
@@ -0,0 +1 @@
+1 1
```
dylanaraps: Alpine versions `ubase` using dates. Current is `20190312`. Though this is rather uncertain as Alpine is the only distributions shipping a non `-git` package for `ubase` on Repology.

https://repology.org/project/ubase/versions

a-schaefers: i suggest we make the version to be the date of the hash we checkout from mforney's repos for both sbase and ubase @dylanaraps

dylanaraps: I'm happy either way. :+1:

a-schaefers: thanks

dylanaraps:
A post install script mentioning the need to add `/opt/bla/bla` to your `PATH` would be nice too.

Example:

```
#!/bin/sh

cat <<EOF

NOTE: This package is installed to /opt/bla/bla.
NOTE: Modify your '$PATH' to include this directory.

-> export PATH="$PATH:/opt/bla/bla"

EOF
```

a-schaefers:
1

> A post install script mentioning the need to add `/opt/bla/bla` to your `PATH` would be nice too.
> 
> Example:
> 
> ```
> #!/bin/sh
> 
> cat <<EOF
> 
> NOTE: This package is installed to /opt/bla/bla.
> NOTE: Modify your '$PATH' to include this directory.
> 
> -> export PATH="$PATH:/opt/bla/bla"
> 
> EOF
> ```
> 
> Edit: Same goes for `sbase` or other `/opt` packages.

We have removed /opt entirely already. There is no /opt

a-schaefers:
ready for merge

PR #134, opened 2019-12-22
new package: dash 0.5.10

a-schaefers:
## Checklist


community/dash/depends
```diff
@@ -0,0 +1 @@
+make
```
dylanaraps: This file can be removed.

community/dash/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+CC="gcc -static" ./configure --prefix=/usr
```
dylanaraps: Change this to the other style. `export CC` and "expanded" `configure`.

```sh
export CC="gcc -static"

./configure \
    --prefix=/usr
```

a-schaefers:
done

community/dash/sources
```diff
@@ -0,0 +1 @@
+http://gondor.apana.org.au/~herbert/dash/files/dash-0.5.10.tar.gz
```
dylanaraps:
```sh
http://gondor.apana.org.au/~herbert/dash/files/dash-0.5.10.2.tar.gz
```

This looks to be the latest version. :+1:

community/dash/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+export CC="gcc -static"
```
konimex: Can this be removed? Since Clang exists in `extra` and is a valid `$CC` option.

dylanaraps: I agree though we'd need to use another method of making builds static.

konimex: `-static` is a `CFLAGS` option so it should be `export CFLAGS="$CFLAGS -static"` should one insist on making a static package.

dylanaraps: Lovely. I'm fine with this change. :+1:

@a-schaefers we may need to go back and fix any packages I may have merged which also set `CC`. I'm happy to do this myself if you like. It's my fault. :)

a-schaefers: let me see what I can do and get back to you. I'm very stubborn about keeping our base be it busybox/gnucoreutils/sbase/ubase/awk/gawk  etc.. all static tools with musl clib. Didn't think about clang.

community/dash/sources
a-schaefers: k

a-schaefers:
New PRs for the aforementioned already merged. Will  fix the pending ones now. Dash is ready.

PR #139, opened 2019-12-23
new package: gnupg2 2.2.19

a-schaefers:
## Checklist


IslandAlpha:
I get this warning:
`gpg: NOTE: THIS IS A DEVELOPMENT VERSION!
gpg: It is only intended for test purposes and should NOT be
gpg: used in a production environment or with production keys!`

IslandAlpha:
Can't seem to figure it out.
On another note we can replace the 'mv's at the end of the script with --enable-gpg-is-gpg2

a-schaefers:
i've been using this package heavily for two weeks without issue (signing git commits, encrypt/decrypt files) I too can't figure out what the warning is about. Our gpg version here is the same as Arch Linux at present. I'm not worried about it.

a-schaefers:
> Can't seem to figure it out.
> On another note we can replace the 'mv's at the end of the script with --enable-gpg-is-gpg2

nice catch thanks

community/gnupg2/build
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -e
+
+# fake
+export MAKEINFO=true
+touch fig2dev convert ; chmod +x fig2dev convert
+export PATH=$PWD:$PATH
```
dylanaraps: Why is this needed? I'll look into whether or not we can fix this a different way.

community/gnupg2/sources
```diff
@@ -0,0 +1 @@
+https://github.com/gpg/gnupg/archive/gnupg-2.2.19.tar.gz
```
dylanaraps: This source should allow you to remove the need for `autoreconf`. 
```
https://gnupg.org/ftp/gcrypt/gnupg/gnupg-2.2.19.tar.bz2
```

community/gnupg2/build
a-schaefers: I'm trying to avoid forcing the user to compile texinfo and friends. If there's another way than how I'm doing it of course I'd be game.

community/gnupg2/sources
a-schaefers: done

community/gnupg2/build
```diff
@@ -0,0 +1,15 @@
...
+
+./configure --prefix=/usr \
+            --sbindir=/usr/bin \
+            --libexecdir=/usr/lib/gnupg \
+            --enable-maintainer-mode \
+            --enable-symcryptrun \
+            --enable-gpg-is-gpg2
```
dylanaraps: Also add `--disable-ldap` and `--disable-nls` (this might also fix the documentation issue).

community/gnupg2/build
dylanaraps: None of the other distributions I've looked at list texinfo as a dependency. Does it build without this workaround?

community/gnupg2/build
konimex: Also `--enable-maintainer-mode` isn't enabled in both Void and Alpine.

a-schaefers: let me take a wack at this. IIRC i built this one based on arch pkgbuild

community/gnupg2/build
a-schaefers: it did not build without the workaround. can't remember if it was texinfo or something other. I will try the new flags below and report back. Basically trying to avoid compiling gigabytes of extra crud, though.

community/gnupg2/build
dylanaraps: Arch does things weirdly sometimes. They tend to enable all optional dependencies too. I like to use CRUX, Adelie, Alpine or Void as references when needed (I look at all 4 and cherry-pick what I need). 

You could also look at https://repology.org. Basically, I find it better to look at more than one distributions package if a reference is needed. :+1:

a-schaefers: got it.

community/gnupg2/build
konimex: I don't know if it's just my KISS installation or what but gnupg2 in my personal repo doesn't have `MAKEINFO=true` and it built successfully. Do you have any `texinfo`-related packages in your system?

a-schaefers:
should be good now

PR #141, opened 2019-12-23
new package: openvpn 2.4.8

a-schaefers:
## Checklist


community/openvpn/build
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh -e
+
+autoreconf -i
```
dylanaraps: Is there no way to avoid `autoreconf`? ie a release tarball of some kind?

a-schaefers: When make doesn't work without ./configure, and when there is only a configure.ac file I have to run autoreconf to create configure so we can make the magic happen. If there is a better way let me know??

konimex: Generally you'll want to find a better source than directly fetching it from GitHub tags (but still from its official site). Try to check the official OpenVPN site, or find the source from Void's `template` for this package.

community/openvpn/sources
```diff
@@ -0,0 +1 @@
+https://github.com/OpenVPN/openvpn/archive/v2.4.8.tar.gz
```
dylanaraps: Try this link: `https://swupdate.openvpn.org/community/releases/openvpn-2.4.8.tar.xz`

This should remove the `autoreconf` requirement.

a-schaefers: cool!

community/openvpn/build
a-schaefers: thanks

community/openvpn/build
```diff
@@ -0,0 +1,13 @@
...
+    --prefix=/usr \
+    --sbindir=/usr/bin \
+    --enable-iproute2 \
+    --enable-pkcs11 \
+    --enable-plugins \
+    --enable-password-save \
+    --enable-x509-alt-username \
+    --disable-plugin-auth-pam
+
+make DESTDIR="$1" install
```
dylanaraps: Missing standalone `make`.

a-schaefers:
fixed

PR #143, opened 2019-12-25
new package: libxaw3d 1.6.3

a-schaefers:
## Checklist


a-schaefers:
this one is done too now

PR #145, opened 2019-12-25
change build and depends: emacs 26.3

a-schaefers:
## Description of package
Change from GTK to Athena toolkit (emacs is well known to not have a proper gtk implementation and the Athena toolkit is more performant)
## Checklist

### DIFF
--- a/community/emacs/depends
+++ b/community/emacs/depends
@@ -1,4 +1,3 @@
-make
 zlib
 ncurses
 libxaw3d
--- a/community/emacs/depends
+++ b/community/emacs/depends
@@ -1,4 +1,3 @@
 zlib
 ncurses
 libxaw3d
-imagemagick6
--- a/community/emacs/depends
+++ b/community/emacs/depends
@@ -1,5 +1,4 @@
 zlib
 ncurses
 libxaw3d
-libmagick6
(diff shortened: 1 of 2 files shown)


a-schaefers:
decided against the mods i had, going basically vanilla for both Emacses

a-schaefers:
I am unhappy with imagemagick6 overall, it was my hope in the beginning it could replace need for maintaining so many image libs, but imagemagick just requires those same libs anyways. let's just use defaults for images

dylanaraps:
Sure thing. I'd much rather we package ImageMagick (7) and drop 6 too. :+1:

community/emacs/depends
```diff
@@ -1,2 +1,4 @@
-ncurses
 zlib
+ncurses
+libxaw3d
+imagemagick6
```
dylanaraps: Does the dependency list need updating too?

a-schaefers:
i dont even want to maintain it. you can drop it

community/emacs/depends
a-schaefers: uhhhhhhh maybe

dylanaraps:
Done :)

PR #149, opened 2019-12-26
new package: esetroot 2003422

a-schaefers:
## Checklist


community/esetroot/depends
```diff
@@ -0,0 +1,2 @@
+make
```
dylanaraps: `make` can be removed as it is an implicit dependency. Same as not listing `gcc`, `musl` etc :)

a-schaefers:
fixed

PR #151, opened 2019-12-27
new package: qrencode 4.0.2

IslandAlpha:
## Description of package

A fast and compact library for encoding data in a QR Code symbol.

## Checklist


dylanaraps:
:+1:

PR #170, opened 2019-12-31
new package: timezone 2019c

IslandAlpha:
## Description of package

IANA time zone database and utilities.

[IANA database page](https://www.iana.org/time-zones)
[github link](https://github.com/eggert/tz)

## Checklist


dylanaraps:
This is already packaged as `tzdata`.  :+1:

Thanks :)

PR #176, opened 2020-01-01
picom: new package at 7.5

konimex:
## Description of package

A `compton` replacement.

Depends on #175.

## Checklist


a-schaefers:
This is a package I originally wanted, but failed to compile. Would be happy to let you maintain all of the deps required for this and to drop compton if we get picom merged.

dylanaraps:
I'm happy to maintain `compton` and keep both around. :+1:

a-schaefers:
okay thanks

PR #177, opened 2020-01-01
new package: netsurf 3.9

a-schaefers:
## Checklist


a-schaefers:
i don't understand how to make vim only a build time dep, please help me fix this

community/netsurf/depends
```diff
@@ -0,0 +1,6 @@
+curl
+gtk+2
+vim vim
```
dylanaraps: `vim make`

Adding `make` after a dependency specifies it as a build and not runtime dependency. :)

community/netsurf/sources
```diff
@@ -0,0 +1 @@
+http://download.netsurf-browser.org/netsurf/releases/source-full/netsurf-all-3.9.tar.gz
```
dylanaraps: Any `https` source available?

community/netsurf/depends
a-schaefers: oh finally get it now. I read the website section 3 or 4 times, too, maybe it should be reworded, or maybe I need more sleep.

community/netsurf/build
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+make
+make DESTDIR="$1" install
```
dylanaraps: No need to set `--prefix=/usr` with `./configure`?

dylanaraps:
Interesting that `vim` is a make dependency! Know why?

community/netsurf/build
a-schaefers: Correct. This netsurf is very purist in the way they do things, if you haven't tried it btw, you might be impressed!

a-schaefers: actually wait now im confused too

a-schaefers: i have it installed, looks like it's in /usr/bin but i don't know which route it took, e.g. did it get installed through the symlink or proper?

a-schaefers:
> Interesting that `vim` is a make dependency! Know why?

Nooo  idea, it's weird. It uses the "xss" of vim iirc...

community/netsurf/build
dylanaraps:
```sh
make \
    LIBDIR=lib \
    INCLUDEDIR=include \
    PREFIX=/usr

make \
    LIBDIR=lib \
    INCLUDEDIR=include \
    PREFIX=/usr \
    install
```

dylanaraps: Also, enable GTK3 as GTK2 will be dropped from KISS once Firefox drops it.

```
    TARGET=gtk \
    NETSURF_GTK_MAJOR=3 \
```

a-schaefers: okay, thought gtk2 might be more forward looking (lol)

a-schaefers: @dylanaraps what about DESTDIR

dylanaraps:
Builds and works fine? Do we need to do anything special to get working SSL support?

a-schaefers:
I typed kiss b netsurf and it failed build because didn't pull in gtk+3 to build first ends/gtk/Makefile:46: *** Unable to find library for: GTK-2 (gtk+-2.0).  Stop.
make

dylanaraps:
My bad. Target should be: `TARGET=gtk3`. :+1:

a-schaefers:
give me a bit to compile gtk+3 before i can test this, because I just purged it a second ago, and realized now i have no cached pkg of it

konimex:
`vim` provides `xxd`. Fortunately it's a `make` depends.

a-schaefers:
> `vim` provides `xxd`. Fortunately it's a `make` depends.

No kidding from a guy who uses Emacs as his login shell in /etc/passwd

a-schaefers:
build failed again https://termbin.com/scfj

dylanaraps:
Arch sets:

```sh
   NETSURF_GTK_MAJOR=3 \
    TARGET=gtk \```

For both `make` and `make install`.

**Comment by a-schaefers** [comment]
> Arch sets:
> 
> ```shell
>    NETSURF_GTK_MAJOR=3 \
>     TARGET=gtk \
> ```
> 
> For both `make` and `make install`.

ok but if you look at the output i do not believe it made it to the second make.

**Comment by a-schaefers** [comment]
okay now it has a new error, this is good https://termbin.com/7k99

**Comment by dylanaraps** [comment]
```
 COMPILE: build/Linux-gtk3/messages_gresource.c
    LINK: nsgtk3
```

It's using GTK3 now at least.

Looks like a simple `sed` command may be needed to fix the `install` call.

a-schaefers:
I think I got it now

a-schaefers:
A missing slash before install, you see it?

dylanaraps:
I do. :)

a-schaefers:
Gtk3 is ugly man i lose that classy look

a-schaefers:
Do we even have any themes or icons in the repos

a-schaefers:
build success ready for merge

dylanaraps:
We do not. I personally don't use any themes so I haven't packaged any.

a-schaefers:
i dont use any either, gtk2 looks better tho imho ;)

PR #187, opened 2020-01-02
socat: new package at 1.7.3.3

aaronNGi:
## Description of package

socat - Multipurpose relay (SOcket CAT)

## Checklist


community/socat/sources
```diff
@@ -0,0 +1 @@
+http://www.dest-unreach.org/socat/download/socat-1.7.3.3.tar.gz
```
dylanaraps: No `https` source available?

aaronNGi: Apparently they use a self-signed cert. We could also use https://repo.or.cz/socat.git/snapshot/e1bc39a53f808f4fdaeb3743c415da861789f351.tar.gz which is the snapshot of 1.7.3.3. What do you think?

dylanaraps: `http` is fine if no `https` is available as we also have the checksum verification to fallback on. :+1:

community/socat/files
```diff
@@ -0,0 +1 @@
+socat-1.7.3.3.tar.gz
```
dylanaraps: This file can be removed entirely.

aaronNGi: Oops, brainfart. Fixed.

dylanaraps:
Thanks :+1:

PR #194, opened 2020-01-05
new package: sinit 1.0

a-schaefers:
## Description of package
```
$ pstree
sinit-+-ubase-box---runsvdir-+-runsv---syslogd
      |                      |-runsv---udevd---udevd
      |                      `-runsv---dhcpcd
      |-ubase-box---emacsfgdaemon-+-Xorg---{InputThread}
      |                         t  |-emacsclient
      |                           `-pstree
      |-5*[ubase-box---ubase-box]
      `-wpa_supplicant
```
It works.

Offers simple alternative to busybox init and easy way to switch service supervisors using /etc/rc.supervisor file [if exists]. All commands are sinit->dash +sbase-box+ ubase-box from pid1 to the getty. Kept orginality of @dylanaraps's boot and shutdown scripts as close as possible to the original kiss init.

```
cat <<EOF

NOTE:
        This package is installed as /usr/bin/sinit
        Please update your kernel parameters
        accordingly,
        -> e.g. init=/bin/sinit
NOTE:
        Also keep in mind that sinit does not use
        the /etc/inittab file. The default busybox
        runit service supervisor will still be
        initialized. To use a different supervisor,
        place the needed code in /etc/rc.supervisor
        and it will automatically be sourced
        instead of busybox runit.
NOTE:
        The 'poweroff' and 'reboot' commands are
        located in /usr/local/bin . Modify your
        'PATH' accordingly or use alias functions.
        In the future this may change when kiss
        gains slotting ability.
        -> poweroff () { /bin/kill -s USR1 1 ; }
        -> reboot () { /bin/kill -s INT 1 ; }

EOF
```

## Checklist


a-schaefers:
give this one a minute if you're looking at it... found a slight issue

a-schaefers:
Okay fixed.

a-schaefers:
in the future i will consider mod this to use /etc/rc.conf to make it with options to e.g. use a different service supervisor than busybox runit. However, with the work done here, it is trivial now to copy this to an overlay and apply changes.

dylanaraps:
Looks good. :+1:

What I'll do is change the default init scripts to define everything as functions. This way you can `. /usr/lib/init/rc.boot` and call the functions you've left unchanged directly. This should reduce the duplication in code between the scripts and ensure you receive any fixes I make to the default.

What do you think?

a-schaefers:
> Looks good. 
> 
> What I'll do is change the default init scripts to define everything as functions. This way you can `. /usr/lib/init/rc.boot` and call the functions you've left unchanged directly. This should reduce the duplication in code between the scripts and ensure you receive any fixes I make to the default.
> 
> What do you think?

Thank you.

Let's cross this bridge later. Yes I like your idea very much, but it was not as simple as simply using ubase-box mount instead of mount everywhere, while it worked most of the time, I did have to modify a couple of your mount commands where ubase did not support all of the flags you were using.

dylanaraps:
These could be issues we can fix in the default boot scripts. Got some kind of list of issues on my end?

a-schaefers:
> These could be issues we can fix in the default boot scripts. Got some kind of list of issues on my end?

I would recommend diffing the 4 files and taking a look. I have good memory, but not that good. HEH

konimex:
Since `baseinit` is completely init-neutral (it can be executed by `busybox init`, `sinit`, `runit-init`, `sysvinit`, anything really), I reckon it'll be better if the `config.h` from this package points to `baseinit` instead with some of @a-schaefers' changes.

dylanaraps:
I agree with this idea. There's no real reason for `sinit` to use `ubase`, `sbase` and `dash` is there?

a-schaefers:
I disagree on this. Using the [suckless] sinit with all of the [suckless] base makes a lot of sense. Surely we can have init freedom in kiss and not be forced to busybox? :P

dylanaraps:
It is "init freedom" to swap busybox init for sinit. What's more difficult is swapping the busybox suite for something else. 

If you'd like to also swap the system coreutils you can create a secondary repository and maintain a small fork of the busybox package (removing any conflicts with `ubase`, `sbase` and `dash`).

Here's what it'd look like:

```
busybox/
├─ build  -> /var/db/kiss/repo/core/busybox/build
├─ checksums 
├─ depends  -> /var/db/kiss/repo/core/busybox/depends
├─ files/
│  ├─ .config # Make your changes here.
│  ├─ .config-suid # Also make changes here if needed.
│  ├─ acpid.run  -> /var/db/kiss/repo/core/busybox/files/acpid.run
│  ├─ crond.run  -> /var/db/kiss/repo/core/busybox/files/crond.run
│  ├─ syslogd.run  -> /var/db/kiss/repo/core/busybox/files/syslogd.run
├─ sources  -> /var/db/kiss/repo/core/busybox/sources
├─ version -> /var/db/kiss/repo/core/busybox/version
┘
```

Basically, you only maintain the `checksums` file and the `.config`. Modifying the `.config` to disable the build of `sh` and the utilities provided by `ubase` and `sbase`.

Users could then use your repository, install `busybox`, `sbase`, `ubase`, `sinit` and `dash` and not have any duplicates at all.

Any incompatibilities in the init scripts or package manager are portability **bugs** on my end which need to be fixed when found.

I know this isn't a full busybox removal as it provides a lot of extra utilities normally provided by other suites but this exact idea can later be extended to replace any busybox tool with another.

a-schaefers:
I have in mind a full conversion at some point, with ability to remove busybox from base, using a base more like that of [oasis](https://github.com/oasislinux/oasis). Using this PR, I am converted from pid 1 to getty busybox-free. Currently kiss breaks without busybox, but I believe this [sinit PR] could have some future value here.

dylanaraps:
Yup. See my above reply if you haven't yet. :)

konimex:
I think it's time to revisit https://github.com/kisslinux/kiss/issues/37 if that's the case. I have a very rough implementation here https://github.com/konimex/kiss/commit/22126019b7ad0de461f9a7302ae52d6bf11425f1 but that's about it.

a-schaefers:
Okay, so I see merit in 1. slotted packages and 2. what @dylanaraps has said, that at some point (probably) forking the busybox package will be unavoidable (because going forward I can't just keep PR'ing installation alternatives in to /usr/local/bin !)

I think the answer here is:

"Both."

1.We need package slotting. 2. Also the default sinit package should hopefully use 
 [modified] rc.boot and rc.shutdown from core baseinit package.

a-schaefers:
And @dylanaraps I do apologize, when you asked earlier I should have just posted a diff myself for everybody to see.

a-schaefers:
> Looks good. 
> 
> What I'll do is change the default init scripts to define everything as functions. This way you can `. /usr/lib/init/rc.boot` and call the functions you've left unchanged directly. This should reduce the duplication in code between the scripts and ensure you receive any fixes I make to the default.
> 
> What do you think?

Coming back to ^^^ this, let me see what I can do here, perhaps I may just open a PR to kiss-init, in addition to a new, additional sinit PR here in community (leaving this one here for the archives) and we may move forward in that way.

> Any incompatibilities in the init scripts or package manager are portability bugs on my end which need to be fixed when found.
And  thank you for saying this. I will keep this in mind.

PR #196, opened 2020-01-06
new package: sinit 1.0 [kiss-init compat.]

a-schaefers:
## Description of package
@dylanaraps 
@konimex 
Spent the whole night thinking it over, this is what I got.

Thanks guys :tada: 

```sh

cat <<EOF

NOTE:

        This package is installed as /usr/bin/sinit
        Please update your kernel parameters
        accordingly,

        -> init=/bin/sinit

        Please keep in mind that sinit does not use
        the /etc/inittab file. The default busybox
        runit service supervisor will still be
        initialized. To use a different supervisor,
        edit /etc/sinit.conf and place script(s) in
        the /etc/sinit.d/ directory suffixed with
        *.sh

        To shutdown:

        -> # kill -s USR1 1

        Advanced users may place in /etc/sinit.d/
        a "rc.shutdown" file which will be sourced
        instead of /lib/init/rc.shutdown

        To reboot:

        -> # kill -s INT 1

        You may opt to install the above commands
        as scripts in /usr/local/bin and modify it
        to the front of 'PATH'.

        In the future this [lack of poweroff and
        reboot scripts in /bin] may change if kiss
        gains slotting abilities or if busybox
        becomes optional. Another possibility is
        a suckless overlay with a forked busybox.

EOF
```

## Checklist


community/sinit/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+make \
+    CFLAGS="$CFLAGS -static"
```
dylanaraps: Feel free to put this on one line ala `make DESTDIR="$1"`. :+1:

dylanaraps:
I wonder if it's a better idea to add the ability to "hook" into the default init script. Maybe the ability to load additional scripts from a directory. This way you can point `sinit` directly to `/usr/lib/init/rc.boot` and pop a script somewhere for the specific changes.

Thoughts?

community/sinit/build
a-schaefers: ok

a-schaefers:
> I wonder if it's a better idea to add the ability to "hook" into the default init script. Maybe the ability to load additional scripts from a directory. This way you can point `sinit` directly to `/usr/lib/init/rc.boot` and pop a script somewhere for the specific changes.
> 
> Thoughts?

If I'm not mistaken, you just read my mind. Take a look at the latest commit!

a-schaefers:
> I wonder if it's a better idea to add the ability to "hook" into the default init script. Maybe the ability to load additional scripts from a directory. This way you can point `sinit` directly to `/usr/lib/init/rc.boot` and pop a script somewhere for the specific changes.
> 
> Thoughts?

I see what you mean. If you look at my latest changes, sinit is now completely modular and offers the user complete and total control over the init process from pid 2 onward, while still using /lib/init/rc.boot by default and not duplicating any code.

I would like to see this merged, however, and if you add hook features to /lib/init/rc.conf , we could see about modifying the sinit package to take advantage of the latest features. How does that sound?

a-schaefers:
I also don't understand a need to force config.h to begin with /lib/init/rc.boot

We can let each init system do what is best for itself I think, in a sane and common sense kind of way, without forcing /lib/init/rc.boot to accommodate everybody.

a-schaefers:
I mean I guess we could rewrite /lib/init/rc.boot to do this stuff and point sinit to it

We could also have /etc/rc.conf and /etc/rc.d in a way similar to what I'm doing here in this PR.

:question:

a-schaefers:
Yeah don't merge this I have something better cooking

PR #200, opened 2020-01-07
New Package: sinit 1.1

a-schaefers:
## Description of package
Depends on https://github.com/kisslinux/kiss-init/pull/3
## Checklist


cemkeylan:
Isn't the latest upstream version 1.1?

a-schaefers:
If that is the case, then yes please do send that email and see if we can get a newer tarball. Thanks @cemkeylan

cemkeylan:
No problem! I have mailed Dimitris (author of sinit) and asked if he could do it.

a-schaefers:
Thanks

community/sinit/build
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh -e
+
+# checking for rc.d support in rc.boot
+if ! grep -q "rc.conf" /lib/init/rc.boot ; then
+    cat << EOF && exit 1
+
+        -> Please update to latest baseinit
+        -> and then install sinit... Thank you.
+EOF
+fi
```
dylanaraps: I'll update `baseinit` before I merge this PR so this can be removed.

dylanaraps:
Just a random thought. I wonder if a generic `inittab` parser would solve the getty issue.

community/sinit/build
```diff
@@ -0,0 +1,21 @@
...
+make CFLAGS="$CFLAGS -static"
+
+install -Dm 755 sinit "$1/usr/bin/sinit"
+
+install -Dm 755 sinit.runit.boot "$1/etc/rc.d/sinit.runit.boot"
+install -Dm 755 sinit.getty.boot "$1/etc/rc.d/sinit.getty.boot"
+install -Dm 755 sinit.pre.shutdown "$1/etc/rc.d/sinit.pre.shutdown"
+install -Dm 755 sinit.post.shutdown "$1/etc/rc.d/sinit.post.shutdown"
+
+if ! grep -q "SINIT_" /etc/rc.conf ; then
```
dylanaraps: This looks a little nicer to me. What do you think?:

```sh
grep -q SINIT_ /etc/rc.conf || cat \

<<EOF >>/etc/rc.conf
# Uncomment and Set to 1 to enable /etc/rc.d/ scripts for sinit
# SINIT_ENABLE_SHUTDOWN_HOOKS=1 # <- (required)
# SINIT_ENABLE_GETTY=1          # <- Enable getty/login   (required)
# SINIT_ENABLE_RUNIT=1          # <- Enable busybox-runit (optional)
EOF
```

community/sinit/files/etc/rc.d/sinit.getty.boot
```diff
@@ -0,0 +1,10 @@
+suckless_getty () {
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty1 linux' &>/dev/null &
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty2 linux' &>/dev/null &
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty3 linux' &>/dev/null &
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty4 linux' &>/dev/null &
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty5 linux' &>/dev/null &
+    sh -c 'ubase-box respawn ubase-box getty /dev/tty6 linux' &>/dev/null &
```
dylanaraps: This could simply be:

```sh
for getty in 1 2 3 4 5 6; do
    ubase-box respawn ubase-box getty "/dev/tty${getty}" linux >/dev/null 2>&1 &
done
```

a-schaefers:
> Just a random thought. I wonder if a generic `inittab` parser would solve the getty issue.

IMO inittab is a sysv remnant we don't need. I prefer this modular setup. It's more flexible.

community/sinit/build
a-schaefers: I used it, but your  <<EOF lacked indent, and we need an empty newline at top of the heredoc

community/sinit/files/etc/rc.d/sinit.getty.boot
a-schaefers: Very concise. I like it

community/sinit/files/etc/rc.d/sinit.runit.boot
```diff
@@ -0,0 +1,5 @@
+busybox_runit () {
+    sh -c 'ubase-box respawn /usr/bin/runsvdir -P /var/service' &
```
dylanaraps:
```sh
ubase-box respawn /usr/bin/runsvdir -P /var/service &
```

a-schaefers: @dylanaraps  we might run into job control issues with getty and this by dropping sh -c. This needs testing.

a-schaefers: I've seen weird things happen when messing with these parts. Has to to do with the way we fork the process

a-schaefers: ok it's tested this is fine

a-schaefers:
So the issue right now is that the parse /etc/rc.conf doesn't parse

cemkeylan:
Okay, so I couldn't reach to Dimitris, but I have uploaded the tarball generated with `make dist` command to <https://dl.carbslinux.org/distfiles/sinit-1.1.tar.gz> . Feel free to use it if you are okay with an external resource. It is not hard to manually check either since it is a minimal codebase.

I have also managed to get it to work on Carbs, and added it to the repositories just now. Keep in mind that I have seperate packages for busybox and busybox-init.

Also, I have comments regarding `baseinit` and the build of sinit here. 

For sinit build:

I don't think we should be editing /etc/rc.conf from build. We should be doing it from post-install. The reasoning is, you do not need to build as root, but you need to install as root. If someone builds sinit without being root, it will fail because of not having write access to /etc/rc.conf.

For baseinit:

Maybe this is some dumb thing I did on my hand, but, run_hooks function is nice, but sinit runs the shutdown command with poweroff|reboot as `$1`. Since run_hooks is a function, it cannot receive the `$1`. I solved this by removing the brackets around and typing out the function. So instead of

```sh
log "Running shutdown pre hooks..."; {
    run_hooks /etc/rc.d/*.pre.shutdown
}
```

it should probably be

```sh
log "Running shutdown pre hooks..."
for file in /etc/rc.d/*.pre.shutdown ; do
    [ -f "$file ] && . "$file"
done
```

a-schaefers:
If I'm not mistaken it should probably just be 

```sh
log "Running shutdown pre hooks..."; {
    run_hooks /etc/rc.d/*.pre.shutdown "$@"
}
```

but needs testing

a-schaefers:
@cemkeylan the run_hooks issue should be a bug filed under kiss-init . Nice catch, though. I will just patch 1.0 to 1.1 so we don't rely on non-authoritative sources. Thanks for offering the hosting though.

cemkeylan:
The thing is when you run `run-hooks` as a function, `$1` becomes `/etc/rc.d/sinit.pre.shutdown' (considering the first file that matches is sinit)

On 13 January 2020 22:22:14 GMT+03:00, Adam Schaefers <notifications@github.com> wrote:
>If I'm not mistaken it should probably just be 
>
>```sh
>log "Running shutdown pre hooks..."; {
>    run_hooks /etc/rc.d/*.pre.shutdown "$@"
>}
>```
>
>but needs testing
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/kisslinux/community/pull/200#issuecomment-573827045

-- 
Cem Keylan

a-schaefers:
Patched to 1.1 about to push the diffs. Easy enough.

community/sinit/build
```diff
@@ -0,0 +1,21 @@
...
+install -Dm 755 sinit.post.shutdown "$1/etc/rc.d/sinit.post.shutdown"
+
+grep -q SINIT_ /etc/rc.conf || cat \
+                                   <<EOF >>/etc/rc.conf
+
+# Uncomment and Set to 1 to enable /etc/rc.d/ scripts for sinit
+# SINIT_ENABLE_SHUTDOWN_HOOKS=1 # <- (required)
+# SINIT_ENABLE_GETTY=1          # <- Enable getty/login   (required)
+# SINIT_ENABLE_RUNIT=1          # <- Enable busybox-runit (optional)
+EOF
```
dylanaraps: This should go in `post-install` as said by @cemkeylan :+1:

community/sinit/files/etc/rc.d/sinit.pre.shutdown
```diff
@@ -0,0 +1,10 @@
+if [ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ]; then
+    case "$1" in
+        reboot|poweroff)
+        ;;
+        *)
+            echo "Invalid action '$1' for rc.shutdown" 1>&2
+            exit 1
+            ;;
+    esac
+fi
```
dylanaraps: What is the point of this file?

cemkeylan: It's from morpheus rc scripts. It did actually save me while I was testing. If your $1 is neither reboot nor poweroff all your services will be stopped but the computer will stay turned on, if this check doesn't happen. Could be removed though.

Post shutdown hook could be changed to

```sh
case "$1" in
        reboot) halt -r ;;
        *) halt -p ;;
esac
```

so that it doesn't require the check on pre shutdown

a-schaefers: It has saved me a couple times, as well, while tinkering.

a-schaefers: > It's from morpheus rc scripts. It did actually save me while I was testing. If your $1 is neither reboot nor poweroff all your services will be stopped but the computer will stay turned on, if this check doesn't happen. Could be removed though.
> 
> Post shutdown hook could be changed to
> 
> ```shell
> case "$1" in
>         reboot) halt -r ;;
>         *) halt -p ;;
> esac
> ```
> 
> so that it doesn't require the check on pre shutdown

Regarding your change on the post shutdown hook, it needs to be guarded 

`if [ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ]; then ...`

because a user could install sinit, without intentions of configuring it immediately, but the post shutdown hook would be running and conflicting with the current busybox configuration.

cemkeylan: Yeah, you are right, I just wanted to keep the comment short, I did not mean to remove the check.

a-schaefers:
09535c2 is necessary because busybox halt does not have -r or -p switches and busybox reboot/poweroff do not work when sinit is PID 1...

dylanaraps:
Shall we swap this to a `git` package and track a hash or would you prefer it as-is?

If you'd prefer it with patches, rename them to `.patch`. :+1:

I forgot all about this.

community/sinit/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+for diff in *.diff; do patch -p1 < "$diff"; done
```
dylanaraps:
```sh
for patch in *.patch; do
    patch -p1 < "$patch"
done
```

community/sinit/files/etc/rc.d/sinit.post.shutdown
```diff
@@ -0,0 +1,10 @@
+if [ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ]; then
```
dylanaraps:
```sh
[ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ] && {
    # code here
}
```

community/sinit/files/etc/rc.d/sinit.pre.shutdown
```diff
@@ -0,0 +1,10 @@
+if [ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ]; then
```
dylanaraps:
```sh
[ "$SINIT_ENABLE_SHUTDOWN_HOOKS" = 1 ] && {
    # code here
}
```

community/sinit/files/etc/rc.d/sinit.runit.boot
```diff
@@ -0,0 +1,5 @@
+busybox_runit () {
+    ubase-box respawn /usr/bin/runsvdir -P /var/service &
+}
+
+[ "$SINIT_ENABLE_RUNIT" = 1 ] && busybox_runit
```
dylanaraps:
```sh
[ "$SINIT_ENABLE_RUNIT" = 1 ] && 
    ubase-box respawn /usr/bin/runsvdir -P /var/service &
```

a-schaefers: I would like to keep the formatting the same across all of the files, how about

```
[ "$SINIT_ENABLE_RUNIT" = 1 ] && {
    ubase-box respawn /usr/bin/runsvdir -P /var/service &
}
```

dylanaraps: :+1:

dylanaraps:
:+1:

PR #204, opened 2020-01-09
Zathura: update to version 0.4.5

tristelune1:
Update of zathura.


dylanaraps:
There's also: zathura-pdf-poppler (I think we call it just poppler).

dylanaraps:
Thanks :+1:

PR #206, opened 2020-01-09
gawk: downngrade to 4.2.1 fixes #192

### LINKED ISSUE #192: bug: gawk causes libgpg-error fail to build
workaround: kiss r gawk && kiss b libgpg-error && kiss i gawk

report upstream: https://dev.gnupg.org/T4459

Apparently there are patches to gawk to fix this, if anyone wants to help

a-schaefers:
## Description of package
https://repology.org/project/gawk/versions

Seems a fairly simple solution to let bigger distros solve this for us in time...
## Checklist


dylanaraps:
Ideally, the best fix would be to patch `gawk` itself. This would fix the issue in all packages.

PR #208, opened 2020-01-09
surfraw: new package 2.3.0

cemkeylan:
## Description of package

surfraw is a basic command-line web search utility.

## Checklist


community/surfraw/build
```diff
@@ -0,0 +1,19 @@
...
+
+# For some reason, surfraw's Makefile
+# gives an error when those directories
+# do not exist (even though the Makefile
+# creates it before exiting with error).
+
+for dir in lib share bin ; do
+	mkdir -p "$1/usr/$dir"
+done
+mkdir -p "$1/etc"
```
dylanaraps: This might be something we're able to fix in a more elegant way. I'll look into it.

dylanaraps: Yup! Setting `MKDIR_P` to run `install` without `-c` fixes the issue.

```sh
#!/bin/sh -e

export MKDIR_P="install -d"

./configure \
    --prefix=/usr \
    --sysconfdir=/etc

make
make DESTDIR="$1" install
```

dylanaraps:
:+1:

PR #209, opened 2020-01-09
new package: argp-standalone 1.3

a-schaefers:
## Description of package
Needed by iucode-tool which is going to be needed by intel-ucode
## Checklist


dylanaraps:
This is probably a better implementation: https://github.com/xhebox/libuargp

It's ripped out of UClibc instead of GlibC and is specifically made to work well with Musl.

a-schaefers:
I will just move all this stuff to my own spot for now. Thanks

PR #211, opened 2020-01-09
new package: intel-ucode 20191115

a-schaefers:
## Checklist


dylanaraps:
This is proprietary so it's better that it live in a personal repository which you could share with others (*see https://github.com/topics/kiss-repo*). I'd be happy for this to live in a repository created by yourself and findable through information in the Wiki.

PR #216, opened 2020-01-12
oed: new package at 6.6

aaronNGi:
## Description of package

Portable OpenBSD ed(1) editor. **Because all operating systems deserve a good editor.**

## Checklist


dylanaraps:
:+1:

PR #218, opened 2020-01-13
neatvi: new package at 07

aaronNGi:
## Description of package

A small vi/ex editor for editing bidirectional UTF-8 text http://litcave.rudi.ir/

## Checklist


community/neatvi/build
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+sed -i -e '/CFLAGS = /c\' -e 'CFLAGS += -Wall' Makefile
```
dylanaraps: Any reason for this line?

community/neatvi/build
```diff
@@ -0,0 +1,7 @@
...
+sed -i s/0600/0666/ ex.c
```
dylanaraps: Same goes for this one.

community/neatvi/build
```diff
@@ -0,0 +1,7 @@
...
+
+make
+install -Dm 755 -- vi "$1"/usr/bin/neatvi
```
dylanaraps: `--` can be removed.

community/neatvi/build
aaronNGi: It's overwriting the global CFLAGS otherwise, no?

community/neatvi/build
aaronNGi: `neatvi` creates files with mode 0600 by default, I don't like it being so strict. Do you want comments for things like this?

community/neatvi/build
dylanaraps: You are correct, though it's also adding `-Wall`?

Could this not be: `sed -i 's/CFLAGS =/CFLAGS +=/' Makefile`?

community/neatvi/build
dylanaraps: Omitting comments is fine so long as there's a written explanation *somewhere* (Comments on this PR for example). This addition may also be temporary if the developer changes it to `0666`.  :+1:

community/neatvi/build
aaronNGi: No, the original Makefile also sets `-O2`. Appending that to `CFLAGS` overwrites the users optimization level. My first idea was to simply remove that entire line but thought maybe it's better to retain warnings. Two substitutions, one to change `=` to `+=`, and one to remove `-O2` would have worked as well but I found this prettier.

community/neatvi/build
aaronNGi: Got it.

Also asked the author to make that change.

dylanaraps:
Thanks :+1:

PR #220, opened 2020-01-15
opendoas: new package at 6.6

ioraff:
## Description of package
Much simpler alternative to `sudo`. Compatibility provided in simultaneous PR to [kiss](https://github.com/kisslinux/kiss) and provided `doas.conf`.

## Checklist


community/opendoas/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+printf "/chown \${BINOWN}:\${BINGRP}/d\n/chmod \${BINMODE}/d\nw\nq" |
+    ed bsd.prog.mk >/dev/null
```
dylanaraps: Rationale behind this?

community/opendoas/build
```diff
@@ -0,0 +1,13 @@
...
+
+./configure \
+    --enable-static \
+    --without-pam
```
dylanaraps: What functionality is lost when `pam` is disabled (if any)?

community/opendoas/files/doas.conf
```diff
@@ -0,0 +1,6 @@
+### Select one of the following. Deleting these may cause package installation
+### and removal to function unexpectedly
+
+permit setenv { KISS_FORCE } :wheel cmd kiss
+#permit persist setenv { KISS_FORCE } :wheel cmd kiss
+#permit nopass setenv { KISS_FORCE } :wheel cmd kiss
```
dylanaraps: Explanation of these three lines?

Also note: The package manager supports other environment variables such as `KISS_PATH`, `KISS_PID`, etc.

ioraff: permits kiss to be run as root if the user belongs to `wheel`, maintaining KISS_FORCE from the invoking user's environment. `persist` caches the privilege escalation for some indeterminate amount of time, and `nopass` allows it to be run without a password. I just commented these out in the same vein as e.g. /etc/sudoers.

Regarding environmental variables, `KISS_FORCE` seems to be the only one used when privileges are escalated (under installation and removal), hence KISS_FORCE="$KISS_FORCE" on lines 969 and 971, with the other environmental variables used only for unprivileged actions such building, etc., but please correct me if I'm wrong.

community/opendoas/build
ioraff: nothing that matters in a KISS system. `shadow` is still enabled by default.

community/opendoas/files/doas.conf
dylanaraps: :+1:

community/opendoas/build
ioraff: no real reason, pretty much just copied over from void, with execute-only permissions assigned in `post-install`.

community/opendoas/files/doas.conf
ioraff: I've just learned that `persist` requires an additional `configure` option due to it being an unstable feature on Linux, so I've removed that commented-out line

community/opendoas/build
dylanaraps: Would this work without the `post-install` or this `ed` call?

ioraff: yes

dylanaraps: So, they could both be removed?

ioraff: ... done

community/opendoas/files/doas.conf
```diff
@@ -0,0 +1,5 @@
+### Select one of the following. Deleting these may cause package installation
+### and removal to function unexpectedly
+
+permit setenv { KISS_FORCE } :wheel cmd kiss
+#permit nopass setenv { KISS_FORCE } :wheel cmd kiss
```
dylanaraps: No more `persist`?

ioraff: I've just learned that it's potentially dangerous on Linux as it requires a different implementation than OpenBSD

dylanaraps: Understood. :+1:

dylanaraps:
Thanks :+1:

PR #224, opened 2020-01-15
2bwm: new package at 0.3.0

icyphox:
## Description of package
A fast floating WM written in XCB, with the particularity of having 2 borders.

## Checklist


community/2bwm/version
```diff
@@ -0,0 +1 @@
+0.3.0 1
```
dylanaraps: Version should be `0.3`.

icyphox: Should be fine now.

dylanaraps:
:+1:

community/2bwm/version
dylanaraps: Thanks :)

PR #226, opened 2020-01-15
opendoas: fix checksum, build

ioraff:
so I was wrong about the edits (are required when building as non-root) and borked a checksum. apologies.


dylanaraps:
:+1:

PR #228, opened 2020-01-16
kakoune: update to version 2020.01.16

Jackojc:
## Checklist


dylanaraps:
:+1:

PR #231, opened 2020-01-17
improve emacs' gnutls settings

a-schaefers:
## Description of package
1. I have spent a lot of time trying to understand the current Emacs security situation, and to the best of my understanding, Emacs development teams favor gnutls to be the default tls, and all other implementations inside Emacs have now gone unmaintained (probably due to being non-gpl licensed), and coming up in Emacs 27 this will be the case even more so. I'm ensuring that our Emacs builds are building explicitly with gnutls support.

2. Improve startup times by ~1 second by decreasing garbage collection frequency to every 50 MB instead of the default 0.76MB (Emacs is optimized for very old computers by default) AND by toggling off package.el (can be turned back on by the user easily, no point in penalizing all Emacs users with a slow startup by having using package.el by always-on by default...) 

## Checklist

### DIFF
--- a/community/emacs/build
+++ b/community/emacs/build
@@ -9,17 +9,25 @@
     --without-dbus \
     --without-gconf \
     --without-gsettings \
-    --with-xpm=no
+    --with-xpm=no \
+    --with-gnutls=yes
 
 mkdir -p "$1/usr/share/emacs/site-lisp"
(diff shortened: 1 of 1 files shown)


dylanaraps:
Depends file needs to mention `gnutls`?

a-schaefers:
I believe it is already mentioned in this one case

PR #233, opened 2020-01-17
sic-git: new package at git

cemkeylan:
## Description of package

Let's celebrate the upcoming kiss feature with the first git package.
sic is "Simple IRC Client". Latest release was made on 2013 and it doesn't build
on musl. The git source tree does build on musl.

## Checklist


dylanaraps:
This needs new checksums. :+1:

cemkeylan:
Added the checksums.

PR #238, opened 2020-01-19
neatvi: update to latest git

aaronNGi:
File creation mode is now in conf.h as `MKFILE_MODE` (still 0600 by default).
New `hll` option to highlight the current line.

## Checklist


dylanaraps:
Nice :+1:

PR #242, opened 2020-01-20
new package: strace 4.1

a-schaefers:
## Checklist


community/strace/depends
```diff
@@ -0,0 +1 @@
+gawk make
```
dylanaraps: Is `gawk` really needed?

a-schaefers:
fixed

PR #243, opened 2020-01-21
gnupatch: new package at 2.7.6

cemkeylan:
## Description of package

gnupatch is nice if you are doing complex patching operations. It is faster
and knows what `/dev/null` is.

## Checklist


community/gnupatch/sources
```diff
@@ -0,0 +1 @@
+http://ftp.gnu.org/gnu/patch/patch-2.7.6.tar.xz
```
dylanaraps: No https?

cemkeylan: Oh sorry must have forgotten

cemkeylan:
Fixed it

a-schaefers:
we already have gnupatch and it is called "patch"

cemkeylan:
I have also replied on the IRC, but what is your point, exactly @a-schaefers ?

PR #246, opened 2020-01-21
9base: new package at 63916da

cemkeylan:
## Description of package

9base is a collection of Plan9 tools by suckless.org. It is much smaller than plan9port
as only the core utilities of Plan9 is installed.

## Checklist


cemkeylan:
I versioned it with the latest commit date, should I change that?

community/9base/version
```diff
@@ -0,0 +1 @@
+20190911 1
```
dylanaraps: The version should be the commit hash.

community/9base/sources
```diff
@@ -0,0 +1 @@
+git+git://git.suckless.org/9base#63916da
```
dylanaraps: I don't know if we should prefer the longer hash or not.

community/9base/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+make
+make DESTDIR="$1" install
+
+# Since rc shell does not conflict with the
+# system tools, link it to /usr/bin
+mkdir -p "$1/usr/bin"
+ln -s /usr/local/plan9/bin/rc "$1/usr/bin/rc"
```
dylanaraps: This can be removed.

cemkeylan:
:+1:

PR #247, opened 2020-01-21
gnupatch: remove duplicate package

cemkeylan:
The package I added turned out to be the duplicate of patch, which is maintained by @a-schaefers . I apologize for missing it before adding the package


dylanaraps:
This is my fault. :)

PR #260, opened 2020-01-23
New package: tdb at 1.4.3

tristelune1:
## Description of package

A Trivial Database similar to GDBM but allows simultaneous commits. It's a dependency to `fdm`. 

## Checklist


tristelune1:
Wrong path. Closing.

PR #261, opened 2020-01-23
pcre: add unicode properties support and fix some perms

ioraff:
`zathura` links against `pcre` when it's present and broke with the glib update. this is the fix.


dylanaraps:
Thanks https://github.com/kisslinux/community/commit/9c4393a1fdfb9f9e0ad6b56465c237d12309d601

PR #262, opened 2020-01-23
New package: fdm at 2.0

tristelune1:
## Description of package

Software to fetch and deliver mail. 

## Checklist


dylanaraps:
Is this patch really needed?

tristelune1:
Yes: [reported issue](https://github.com/nicm/fdm/issues/71)

Because the patch modify `Makefile.am`, it was necessary to add `automake` as a dependency. With the next release, I should be able to simplify the `build` file. 

But probably the next release won't come soon.

tristelune1:
Unfotunately it didn't work. It I understand your idea, I ran `aclocal` and `automake` and did a `diff` on the `Makefile.in`. 

You can have a look at my new patch. I can say it compiles and I have no error if I execute `fdm`. But I didn't test it extensively.

community/fdm/build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+patch -p1 < musl.patch
+
+./configure \
+	--prefix=/usr \
+	--enable-pcre
```
dylanaraps: Indentation is 4 spaces.

dylanaraps:
OK. I think it then makes more sense to swap back to how you had it at the start. Apologies...

tristelune1:
No problem, I learnt a little bit better the build system by the way. It should be ok now.

community/fdm/build
```diff
@@ -2,9 +2,12 @@
 
 patch -p1 < musl.patch
 
+aclocal
+automake
```
dylanaraps: Can this simply be: `autoreconf` or `autoreconf -i`?

tristelune1: `autoreconf` seems to be enough. Done!

PR #268, opened 2020-01-27
    * Add xtrlock build

dzove855:
## Description of package
Lock X display until password supplied, leaving windows visible

## Checklist


community/xtrlock/build
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+make
+make DESTDIR="$1" install
```
dylanaraps: The Makefile can be removed and the build process replaced with:

```sh
"${CC:-gcc}" -lX11 -DSHADOW_PWD -o xtrlock xtrlock.c

install -Dm4755 xtrlock     "$1/usr/bin/xtrlock"
install -Dm644  xtrlock.man "$1/usr/share/man/man1/xtrlock.1"
```

community/xtrlock/sources
```diff
@@ -0,0 +1,2 @@
+http://deb.debian.org/debian/pool/main/x/xtrlock/xtrlock_2.12.tar.xz
```
dylanaraps: `https` source works. :+1:

PR #269, opened 2020-01-27
osh: bump to 0.7.0

jedahan:
## Checklist


dylanaraps:
:+1:

PR #271, opened 2020-01-29
sbase: use official git sources

a-schaefers:
## Checklist


dylanaraps:
While you're at it, `sbase` should install directly to `/usr/bin/` now. :)

a-schaefers:
If I understand you correctly, it is now okay.

PR #274, opened 2020-01-29
cfm: new package at v0.1.2

WillEccles:
## Description of package
[cfm](https://github.com/willeccles/cfm) (Cactus File Manager) is a simple TUI file manager written in pure C with no dependencies.

## Checklist


dylanaraps:
:+1:

PR #284, opened 2020-02-01
gnupg2: update build for alternatives

a-schaefers:
## Checklist


dylanaraps:
Might want to leave this as-is I think.

a-schaefers:
I think most distros now days have gpg2 is gpg by default and gpg legacy is gpg1 but i could be totally wrong? + with our alternatives I figured it wouldn't be a big deal to let users choose 1 or 2, however with this update I am not sure if you could have both at the same time, for for that matter it's probably best to leave kiss gnupg the way it is, and not break things for anybody. Closing

PR #287, opened 2020-02-01
gtar: update build for alternatives

a-schaefers:
## Checklist


dylanaraps:
Is this compatible with the package manager?

a-schaefers:
yep :)

PR #295, opened 2020-02-02
cfm: update to 0.4.4

WillEccles:
## Description of package
[cfm](https://github.com/willeccles/cfm) is a simple TUI file manager with no dependencies.

## Checklist


community/cfm/version
```diff
@@ -1 +1 @@
-0.4.1 3
+0.4.4 4
```
dylanaraps: `0.4.4 1`. You only bump the second field when you change the package's files but there as no package release.

WillEccles: Could you elaborate on what this means? I'm not understanding properly. Fixed, however.

dylanaraps: - Software has an update -> First field (software version) is bumped.
- The KISS package files were changed (Possibly to change the `build` file or to fix a now invalid source as examples) -> Second field is bumped.

WillEccles: Okay, thanks! Will keep that in mind.

PR #300, opened 2020-02-02
gdb: new package at 8.3.1

icyphox:
Signed-off-by: Anirudh Oppiliappan <x@icyphox.sh>

## Description of package

The GNU Debugger

## Checklist


community/gdb/sources
```diff
@@ -0,0 +1 @@
+http://ftp.gnu.org/gnu/gdb/gdb-8.3.1.tar.xz
```
dylanaraps: https?

icyphox: Woops. Didn't see that. Fixing.

PR #302, opened 2020-02-03
weechat: add ncurses dependency

ndowens:
ncurses was missing from dependency when I attempted to install weechat


dylanaraps:
Fixed, thanks.

PR #310, opened 2020-02-04
slmenu: new package at 0.1

artemkobets:
## Description of package
Dmenu spinoff for the console
## Checklist


dylanaraps:
The `depends` file can be removed if it is empty. :+1:

dylanaraps:
:+1:

PR #276, opened 2020-01-29
sinit: build without depending on ubase

cemkeylan:
Resolves #273 

Okay, so this first commit might be a little rough on the edges. I have added a modified version of the build as packaged in Carbs. I double checked for different directory names (`/etc/init` to `/etc/rc.d`) but I might have something mixed.

I will talk about the rationale of all the changes.

### Adding ubase halt and respawn

This is obviously a given if we don't want to add ubase dependency just for two simple utilities. `busybox halt` is not compatible with `sinit`. Also, sinit doesn't have a built-in respawn function, we probably require `respawn` from ubase.

I am saying "probably" here, because we don't necessarily depend on it. You don't have to restart getty if it exits, but realistically speaking, you would want getty to restart if you exit it. Therefore, it is built-in to the package.

### Adding poweroff and reboot scripts

Replacing init should feel natural. Every init system comes with a complementary `poweroff` and `reboot` tools. Nobody would actually want to write `/bin/kill -s USR1 1` every time they want to shut their computers down.

### Boot and shutdown hooks

I have removed the `pre.shutdown` hook and changed the `post.shutdown`.

```sh
# We handle the shutdown from here

if [ "$SINIT_SHUTDOWN_HOOKS" = 1 ]; then
        case "$1" in
                reboot)
                        out "Requesting system reboot..."
                        halt -r
                        ;;
                *)     # I have changed this from shutdown to *
                        out "Requesting system poweroff..."
                        halt -p
                        ;;
        esac
fi
```
Now that our sinit build is quite stable, we don't need a pre-shutdown check to control the validity of the variable. Just shutdown.

I have also merged the boot-hook onto a single file to avoid crowd.

### Issues

* ~~I have a leftover `# vim:filetype=sh` on `post.shutdown`~~

PR #311, opened 2020-02-05
ufw: new package at 0.36

jlxip:
## Description of package
Uncomplicated Firewall, a program for managing a netfilter firewall. It provides a command line interface and aims to be uncomplicated and easy to use.

## Checklist


community/ufw/depends
```diff
@@ -0,0 +1,3 @@
+python
+iptables
+# 'gettext' is also marked as a dependency, but UFW seems to work without it.
```
dylanaraps: This line can be removed.

community/ufw/build
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
+
+python ./setup.py install --root="$1"
```
dylanaraps: ` python setup.py install --root="$1"`

dylanaraps: Also add (before this line):

```sh
sed -i 's|/lib|/usr/lib|;s|sbin|bin|' setup.py
```

jlxip: I'm sorry. First time. Why is the `sed` replacement required?

dylanaraps: It changes the install location for `ufw` to `/usr/lib` instead of `/lib` and `bin` instead of `sbin`. Not 100% required and feel free to omit it if all works well regardless. :+1:

community/ufw/depends
```diff
@@ -0,0 +1,2 @@
+python
+iptables
```
dylanaraps:
```
iptables
python
```

community/ufw/build
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+sed -i 's|/lib|/usr/lib|;s|sbin|bin|' setup.py
```
dylanaraps: Add a newline below the `sed` line.

jlxip:
I give up :100:

dylanaraps:
Packages in KISS go through a thorough review before they're accepted into the repositories. This goes beyond shellcheck and includes nitpicks about the package format and build process. This is to ensure that every package is of a high quality and the best that it can be.

Once a package is in the repositories however, it's rare that it change beyond a simple version bump. It's better to go through this process before a package is in the repositories rather than after as we may come across issues which need fixing (*install location of ufw as an example*). 

A good example of a package which is seeing a lot of attention, changes and fixes is `sinit`. While it is already in the repositories, it is seeing major changes and fixes to work without `ubase` and better incorporate with KISS' init "framework". (https://github.com/kisslinux/community/pull/276)

All of this being said; packages do get accepted without a single peep from others. As contributors submit packages they learn the review process and are then able to review with their own eyes.

I'd rather you give this another chance and see this PR through. It's a shame to give up. You're welcome to keep this package local to your own repository of course. The choice is yours!

jlxip:
Sorry, I did not explain myself. If anything, I'm thankful to you for giving so much attention to the project and for paying attention to whoever wants to create a package :heart: 

However, the problem is with the package I was trying to make: turns out I actually didn't get it working, and I realised a bit too late (after the pull request had been done). I was getting errors when testing it better (in `ufw enable`) which I've tried to fix but couldn't manage to. I'm not sure if they have to do with the kernel configurations I've tried or if it's an issue with the relations between UFW and KISS' configurations itself.

I'm no expert in distro developing; in fact, this is the first package I tried to create for any distro, but as I couldn't get it to work, I thought I'd just leave it to someone who really knows what they're doing.

PR #316, opened 2020-02-08
xmlsec1: new package at 1.2.29

artemkobets:
## Description of package
XML security library. Dependency for `oath-toolkit`
## Checklist


community/xmlsec1/sources
```diff
@@ -0,0 +1 @@
+http://www.aleksey.com/xmlsec/download/xmlsec1-1.2.29.tar.gz
```
dylanaraps: No `https`?

community/xmlsec1/depends
```diff
@@ -0,0 +1,2 @@
+libressl
```
artemkobets: I'm not sure about dependencies here.
According to FAQ https://www.aleksey.com/xmlsec/faq.html xmlsec can be built against
different crypto libraries (openssl/libressl, gnutls, libgcrypt, nss).
I only added `libressl` to depends file to satisfy minimum requirements, but would it make more sense to include `gnutls`, `libgcrypt` and `nss` here as well?

community/xmlsec1/sources
artemkobets: oops, fixed

community/xmlsec1/depends
dylanaraps: It would not. You should be able to use configure flags to limit it to libressl, if not it is fine as-is. :+1:

artemkobets: Added flags and a missing `libtool` dependency. Should be good now

dylanaraps:
:+1:

PR #317, opened 2020-02-08
oath-toolkit: new package at 2.6.2

artemkobets:
## Description of package
Tools for working with one-time passwords
## Checklist


community/oath-toolkit/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+patch -p1 < intprops.patch
+
+./configure --prefix=/usr
```
dylanaraps:
```sh
./configure \
    --prefix=/usr
```

community/oath-toolkit/depends
```diff
@@ -0,0 +1,2 @@
+pkgconf make
```
dylanaraps: This is an implicit dependency and can be removed (same as `make`, `gcc` for reference).

dylanaraps:
:+1:

PR #321, opened 2020-02-11
dosbox

jedavies-dev:
## Description of package
DOSBox - an x86 emulator with DOS

## Checklist


community/dosbox/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+./autogen.sh
```
dylanaraps: Does this run `autoconf` and friends? If so they should be added to `depends`. :+1:

jedavies-dev: Yes, I did consider that.  So to test I did "kiss r autoconf", and then "kiss b dosbox" again, and it still worked.  So I guess it doesn't need autoconf?

community/dosbox/version
```diff
@@ -0,0 +1 @@
+0.74.3 1
```
jedavies-dev: This is wrong - will fix.

community/dosbox/build
jedavies-dev: Hmm, the script just calls autoconf and friends anyway, so I think it should probably be in the deps list.  Will add.

dylanaraps: Does the script need to be run? If the build works without it, it may just skip calling them if unneeded.

jedavies-dev: It seems that it can work on my system without running autogen.sh first, though perhaps that may not be true for other people? I guess what is guidestone on this... "If autogen.sh is present, don't run it if you don't need to"?

dylanaraps: Yeah. Will add it to the list tomorrow. We avoid running the `auto*` stuff if unneeded.

jedavies-dev: Got it - will  remove running of autogen.sh from build and remove autoconf from deps.

dylanaraps:
:+1:

PR #326, opened 2020-02-12
new package: emacs27-git

a-schaefers:
## Description of package
I am excited about this, with the new pdumper people with bloated configs will still be able to load fast, in addition we can now enable json features which will improve performance with packages that use it, such as the up and coming lsp-mode.
## Checklist


a-schaefers:
@dylanaraps I can't get it to checkout the emacs-27 branch, which clearly exists (one can see it by looking online in a web browser if desired...)

```
-> emacs27-git Extracting sources
-> Checking out emacs-27 
fatal: ambiguous argument 'emacs-27': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
!> Commit hash emacs-27 doesn't exist 
bash-5.0$ 
```

dylanaraps:
Pushed a package manager update which fixes this. :+1:

PR #331, opened 2020-02-12
new package: wget 1.20.3

a-schaefers:
## Checklist


community/wget/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+
+./configure --prefix=/usr
```
dylanaraps:
```sh
./configure \
    --prefix=/usr
```

PR #336, opened 2020-02-12
less: new package at 551

artemkobets:
## Description of package
Couldn't figure out why git diff wouldn't show color in output with busybox less, so here's a bigger alternative.
## Checklist


community/less/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+./configure \
+    --prefix=/usr \
+    --with-regex=pcre
```
dylanaraps: Can this dependency be removed?

artemkobets: Didn't realize it was optional, fixed.

PR #337, opened 2020-02-13
new package: iproute2 5.5.0

a-schaefers:
## Checklist


community/iproute2/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+patch -p1 < 0001-make-iproute2-fhs-compliant.patch
+
+sed -i 's/-Werror//' Makefile
+
+./configure
```
dylanaraps: Missing `--prefix`? Can't `--sbindir` also be set here?

community/iproute2/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+patch -p1 < 0001-make-iproute2-fhs-compliant.patch
+
+sed -i 's/-Werror//' Makefile
```
dylanaraps: Does this line have to exist?

a-schaefers: Nooooooope

community/iproute2/build
a-schaefers: I just copied arch on that, figured it's part of their base so it'd be a quality pkgbuild ...

dylanaraps: Gross: https://github.com/shemminger/iproute2/blob/master/configure

a-schaefers: ok well it works either way

community/iproute2/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+patch -p1 < 0001-make-iproute2-fhs-compliant.patch
+
+./configure \
+    --prefix=/usr \
+    --sbindir=/usr/bin
```
dylanaraps: These don't actually work. Default prefix is `/usr` and you've probably installed these to `/usr/sbin`!

community/iproute2/build
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+
+patch -p1 < 0001-make-iproute2-fhs-compliant.patch
+
+./configure \
+    --prefix=/usr
```
dylanaraps: You can remove this as well. Just `./configure` is fine.

a-schaefers:
SBINDIR is needed for second make. --sbindir=/usr/bin does nothing on configure, and I am unsure if --prefix=/usr does anything useful or not, however I have added it per your last comment, just incase.

a-schaefers:
Ok got it

dylanaraps:
As I said above, they don't do anything ;)

a-schaefers:
merge me baby

PR #345, opened 2020-02-13
nnn: change to nnn-bin and bump to 3.0

monochromal:
## Description of package
:dolphin: The missing terminal file manager for X.
## Checklist


dylanaraps:
I really don't see the point in a `-bin` package for `nnn`. What's wrong with compiling it?

PR #346, opened 2020-02-13
new package: shadow 4.8.1

a-schaefers:
## Checklist


dylanaraps:
`libXslt` can be removed as a depend with `--disable-man`. The same may be doable for `webkitgtk` which is the only package that requires it.

a-schaefers:
dep is fixed now to use libXslt

dylanaraps:
Read my above comment.

a-schaefers:
done

PR #358, opened 2020-02-13
bash: static

a-schaefers:
## Checklist


dylanaraps:
Done.

PR #364, opened 2020-02-14
new package: gc 8.0.4

a-schaefers:
## Checklist


community/gc/depends
```diff
@@ -0,0 +1 @@
+linux-headers make
```
dylanaraps: Remove this entire file.

a-schaefers:
done

PR #370, opened 2020-02-14
tcl: use system sqlite, symlink lib, install macros

a-schaefers:
## Checklist


community/tcl/depends
```diff
@@ -1 +1,2 @@
 zlib
```
dylanaraps: Sort this list.

community/tcl/build
```diff
@@ -1,14 +1,21 @@
 #!/bin/sh -e
 
-IFS=. read -r ver1 ver2 _ < "${0%/*}/version"
+IFS=. read -r maj min _ < "${0%/*}/version"
 
-unix/configure \
+rm -rf pkgs/sqlite3*
```
dylanaraps: This can be removed if `--system-sqlite` is used no?

community/tcl/build
```diff
@@ -1,14 +1,21 @@
...
+./configure \
     --prefix=/usr \
     --mandir=/usr/share/man \
     --enable-threads \
-    --enable-64bit
+    --enable-64bit \
+    --with-system-sqlite
 
 make
-make -j1 INSTALL_ROOT="$1" install install-private-headers
```
dylanaraps: Why was `-j1` removed? Unneeded?

community/tcl/build
```diff
@@ -1,14 +1,21 @@
...
+make INSTALL_ROOT="$1" install install-private-headers
 
-ln -s "/usr/bin/tclsh${ver1}.${ver2}" "$1/usr/bin/tclsh"
+ln -s "/usr/bin/tclsh${maj}.${min}" "$1/usr/bin/tclsh"
+ln -s "/usr/lib/libtcl${maj}.${min}.so" "$1/usr/lib/libtcl.so"
+install -Dm644 tcl.m4 -t "$1"/usr/share/aclocal
```
dylanaraps:
```sh
"$1/usr/share/aclocal"
```

community/tcl/build
```diff
@@ -1,14 +1,21 @@
...
 
-IFS=. read -r ver1 ver2 _ < "${0%/*}/version"
+IFS=. read -r maj min _ < "${0%/*}/version"
 
-unix/configure \
+rm -rf pkgs/sqlite3*
+
+cd unix
+
+./configure \
```
dylanaraps: `unix/configure`? Why change it?

community/tcl/build
ioraff: yes, unneeded

community/tcl/build
ioraff: avoids having to specify `unix/...` multiple times as per the changes in this PR.

dylanaraps: I only see `unix/...` being used once for `configure`?

community/tcl/build
ioraff: based on other distros' build files and the source directory itself, it's a bit hard to tell if `--with-system-sqlite` is still an option in 8.6.10, so this is just forcing it.

community/tcl/build
ioraff: it'd also have to be specified on the last line (if not for the `cd`)

a-schaefers: I went with the cd because my thinking is there is also a win/ and a macos/ directory each with their own makefiles as well, and to cd explicitly to the unix/ directory and work entirely from it seemed a safe choice.

community/tcl/build
a-schaefers: I copied that -j1 from adelie, originally. It's probably because they target so many arches or? It's safe to remove for us, AFAIK. But there are sometimes weird make bugs that require -j1 for *some* and not *all* of us ... @dylanaraps Would you prefer I reintroduce the -j1 ?

community/tcl/build
```diff
@@ -1,14 +1,21 @@
...
     --prefix=/usr \
     --mandir=/usr/share/man \
     --enable-threads \
-    --enable-64bit
+    --enable-64bit \
+    --with-system-sqlite
```
dylanaraps: This can be removed.

community/tcl/build
dylanaraps: There's no need to `cd`.

PR #373, opened 2020-02-14
tcl: fix build

ioraff:
when `--with-system-sqlite` is specified, the build fails when linking `tclsqlite.c` with a bunch of undefined references. deleting `pkgs/sqlite3*` fixes this.
 
## Existing package

### DIFF
--- a/community/tcl/build
+++ b/community/tcl/build
@@ -2,6 +2,8 @@
 
 IFS=. read -r maj min _ < "${0%/*}/version"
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
@a-schaefers

dylanaraps:
I'll also add.

@a-schaefers Test that the changes actually work.

@ioraff I'd prefer if the maintainer was contacted first as they are **solely** responsible for their packages and are the only ones able to push changes.

PR #376, opened 2020-02-14
libmupdf: new package at 1.16.1

ioraff:

## Description of package
mupdf libs and command line tools

## New package

## Existing package


community/libmupdf/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+patch -p1 < pkcs7-fix.patch
+
+export CFLAGS="$CFLAGS -fPIC"
+export CXXFLAGS="$CXXFLAGS -fPIC"
```
dylanaraps: Needed?

community/libmupdf/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+patch -p1 < pkcs7-fix.patch
```
dylanaraps: What does the patch do?

ioraff: I believe it's a libressl fix. Build fails without it.

community/libmupdf/build
ioraff: Not absolutely

PR #377, opened 2020-02-14
zathura-pdf-mupdf: new package at 0.3.5

ioraff:

## Description of package
mupdf frontend to zathura

## New package

## Existing package


community/zathura-pdf-mupdf/depends
```diff
@@ -0,0 +1,9 @@
+cairo
+jbig2dec
+libjpeg-turbo
+libmupdf make
+libressl
+meson make
+openjpeg2
+samurai make
```
dylanaraps: This can be removed.

community/zathura-pdf-mupdf/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+meson build \
+    --prefix=/usr \
+    -Dlink-external=true \
+
+ninja -C build
+DESTDIR="$1" ninja -C build install
```
dylanaraps: See: 

```sh
#!/bin/sh -e

export DESTDIR="$1"

meson \
    --prefix=/usr \
    --sysconfdir=/etc \
    --mandir=/usr/share/man \
    -Ddocs=false \
    -Dintrospection=false \
    . output

ninja -C output
ninja -C output install
```

This is how we typically format `meson` builds.

community/zathura-pdf-mupdf/depends
```diff
@@ -0,0 +1,8 @@
+cairo
+jbig2dec
+libjpeg-turbo
+libmupdf make
```
dylanaraps: Only a `make` dep?

ioraff: Yes. As far as libraries go, `libmupdf` only contains object archives and headers. I've tested this.

dylanaraps: :+1:

PR #387, opened 2020-02-14
transmission-daemon: new package at 2.94

ioraff:

## Description of package
Torrent daemon and CLI remote

## New package

## Existing package


community/transmission-daemon/build
```diff
@@ -0,0 +1,19 @@
...
+    --enable-cli \
+    --enable-daemon \
+    --enable-utp \
+    --without-systemd-daemon \
+    --without-gtk
+
+make CXXFLAGS="$CXXFLAGS -std=c++11"
+make -C daemon DESTDIR="$1" install
+
+rm -f "$1/usr/share/transmission/web/LICENSE"
```
dylanaraps: Remove this.

community/transmission-daemon/build
```diff
@@ -0,0 +1,19 @@
...
+    --sysconfdir=/etc \
+    --localstatedir=/var \
+    --disable-nls \
+    --enable-cli \
+    --enable-daemon \
+    --enable-utp \
+    --without-systemd-daemon \
+    --without-gtk
+
+make CXXFLAGS="$CXXFLAGS -std=c++11"
```
dylanaraps: `make -C daemon` perhaps?

Also, why the `-std=c++11`?

community/transmission-daemon/build
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -e
+
+./configure \
+    --prefix=/usr \
+    --sbindir=/usr/bin \
+    --mandir=/usr/share/man \
+    --sysconfdir=/etc \
+    --localstatedir=/var \
+    --disable-nls \
+    --enable-cli \
```
dylanaraps: Does this install with `make -C daemon intall`?

community/transmission-daemon/depends
```diff
@@ -0,0 +1,3 @@
+pkgconf make
```
dylanaraps: Remove this line.

community/transmission-daemon/build
ioraff: > `make -C daemon` perhaps?
Needs to build libs first, so no.
> Also, why the `-std=c++11`?
Taken from another build, turns out unnecessary.

ioraff: > `make -C daemon` perhaps?

need to build libs first, so no

> Also, why the `-std=c++11`?

turns out unnecessary.

community/transmission-daemon/build
ioraff: removed. `cli` is different from daemon/remote.

PR #397, opened 2020-02-15
Added custard, a window manager by Sweets.

periish:

## New package

## Existing package


community/custard/build
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
```
dylanaraps: Add a newline below the shebang.

PR #414, opened 2020-02-16
vimpc: new package at 0.09.2

cemkeylan:
Read: https://k1ss.org/style.txt
Read: https://k1ss.org/guidestones.txt

---

## New package

## Existing package


community/vimpc/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+./autogen.sh
```
dylanaraps: https://k1ss.org/style [0401]

community/vimpc/depends
```diff
@@ -0,0 +1,4 @@
+automake make
+libmpdclient
+ncurses make
```
dylanaraps: Runtime dep?

cemkeylan: Yeah libmpdclient is a runtime dependency for all mpd clients.

community/vimpc/build
cemkeylan: There aren't any sources for this.

community/vimpc/depends
dylanaraps: I meant `ncurses` being `make`.

community/vimpc/build
dylanaraps:
```
|  If autogen.sh or autoreconf are required, prefer     |
|  autoreconf.                                          |
|               
```

PR #426, opened 2020-02-18
libmupdf: compile with -fPIC, fixing zathura-pdf-mupdf linking

ioraff:
Read: https://k1ss.org/style

---

## New package

- [ ] I agree to maintain this package.
- [ ] I agree to notify @dylanaraps if I can no longer maintain this package.

## Existing package


community/libmupdf/build
```diff
@@ -1,5 +1,8 @@
 #!/bin/sh -e
 
+export CFLAGS="$CFLAGS -fPIC"
+export CXXFLAGS="$CXXFLAGS -fPIC"
+
```
dylanaraps: I'd rather we didn't set CFLAGS like this unless it makes sense (coreutils as static for example). This can be done through your CFLAGS or for finer control, through `KISS_HOOK`.

dylanaraps: Does this fix a compile bug with our normal GCC or your modified one? If the latter, my previous message applies.

ioraff: I'm running our normal GCC. These C{,XX}FLAGS are [necessary](https://bugzilla.redhat.com/show_bug.cgi?id=1109589) regardless, I assume due to the fact that `libmupdf` is static only and `zathura-pdf-mupdf` is a shared object, and so requires all PIC.

dylanaraps: I can reproduce. Must both CFLAGS and CXXFLAGS be set?

ioraff: Turns out just CFLAGS.

PR #433, opened 2020-02-19
new package: readline 8.0p004

a-schaefers:
Read: https://k1ss.org/style

---

## Description of package
Turns out I need this package to make the ruby irb repl function. So I figure others will end up needing this package for something or other sooner or later. Add it to my list!

Note: 8.0p004 or 8.0_p4 or 8.0p4 all seem to be acceptable repology version naming conventions.

Note: I also chose to download the "patches" from the official gnu upstream source, (instead of keeping a local copy), and because they are official patches, I believe this makes sense.

## New package

## Existing package


community/readline/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+for patch in readline80-00* ; do patch -p0 < "$patch"; done
```
dylanaraps: https://k1ss.org/style.txt 1800

community/readline/build
```diff
@@ -0,0 +1,8 @@
...
+
+./configure \
+    --prefix=/usr
+
+make DESTDIR="$1" install
```
dylanaraps: https://k1ss.org/style 0400

community/readline/build
a-schaefers: In this case the files are staying on the upstream official source, I would like to request a style guide exception.

dylanaraps: I meant breaking it onto multiple lines. See 1800.

a-schaefers: oh, okay

a-schaefers: I thought it was the glob, the problem

PR #435, opened 2020-02-19
xkb-switch: new package at 1.6.0

artemkobets:
Read: https://k1ss.org/style

---

## Description of package
Query and change the XKB layout state.

## New package

## Existing package


community/xkb-switch/build
```diff
@@ -0,0 +1,13 @@
...
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=/usr
+
+cmake --build   build
+cmake --install build
+
+# man page may be compressed, install manually instead
+rm -rf "$1/usr/share/man"
+install -Dm 0644 man/xkb-switch.1 "$1/usr/share/man/man1/xkb-switch.1"
```
artemkobets: this is a workaround due to a cmake rule that compresses man page if it finds xz or gzip installed.

community/xkb-switch/build
```diff
@@ -0,0 +1,13 @@
...
+export DESTDIR="$1"
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=/usr
+
+cmake --build   build
+cmake --install build
+
+# man page may be compressed, install manually instead
+rm -rf "$1/usr/share/man"
```
dylanaraps: Won't the `install` command overwrite the man page anyway?

community/xkb-switch/build
```diff
@@ -0,0 +1,13 @@
...
+
+export DESTDIR="$1"
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=/usr
+
+cmake --build   build
+cmake --install build
+
+# man page may be compressed, install manually instead
```
dylanaraps: https://k1ss.org/style 0205

community/xkb-switch/build
dylanaraps: I just checked the source, this can't be turned off either. See my other comment. :+1:

community/xkb-switch/build
artemkobets: man page is installed with .xz or .gz extension, does not get overridden because of this.

dylanaraps: You're right. :+1:

community/xkb-switch/build
```diff
@@ -0,0 +1,13 @@
...
+
+export DESTDIR="$1"
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=/usr
+
+cmake --build   build
+cmake --install build
+
+# Man page may be compressed, install manually instead
```
dylanaraps: Full stop (`.`) and will merge. :+1: 

Thanks

PR #438, opened 2020-02-20
uemacs: add new package at git to community

cemkeylan:
Read: https://k1ss.org/style

---

## Description of package

Git package of Micro Emacs. This is a really portable and small implementation of emacs (less than 200kb in size) currently maintained by Linus Torvalds.

**Rationale for not using `make install`:**

* It does not create directories
* It installs emacs.rc as a hidden file to `/usr/lib` for some reason

## New package

## Existing package


community/uemacs/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+make
```
dylanaraps: Newline below make.

Issue #415, opened 2020-02-17
More packages up for adoption: (I will no longer be maintaining these...)


a-schaefers:
compton
~~libxdg-basedir (required by compton & picom)~~ `konimex`
~~libconfig (required by compton & picom)~~ `konimex`

~~netsurf~~ `IslandAlpha`
~~perl-html-parser (req. by netsurf)~~ `IslandAlpha`
~~perl-html-tagset (req. by netsurf)~~ `IslandAlpha`

esetroot

keychain

konimex:
I'll take both `libxdg-basedir` and `libconfig`.

IslandAlpha:
I'll take netsurf and its deps

PR #454, opened 2020-02-24
Various packages: adoption (#415)

IslandAlpha:
Adopting netsurf, perl-html-parser, and perl-html-tags.

Changing email on apulse and qrencode.

## New package


dylanaraps:
Is there any purpose to `apulse` if you don't mind me asking? None of the packages in the repositories require it no?

IslandAlpha:
I added it for firefox since it only supports pulseaudio

dylanaraps:
No no. Our Firefox works with ALSA. There's no need for apulse or pulseaudio. In reality, Mozilla just compile their Firefox builds with `--disable-alsa`.

IslandAlpha:
So they do `--disable-alsa` and tell everyone they need to use poetterware for audio? Wow.

In that case there's no need for `apulse`, lets get rid of it

dylanaraps:
Pretty much. They say they won't maintain it in the codebase anymore though I haven't seen any movement over the years to remove the code itself.

Awesome. Shall I remove it for you?

IslandAlpha:
Go ahead

PR #459, opened 2020-02-25
libxdg-basedir: drop package

konimex:
The only package "depending" on libxdg-basedir is `compton`, and since @a-schaefers is dropping maintenance of compton (and as of the time when this PR was submitted no one has an interest to maintain it) I suggest we drop compton too.

## Existing package


dylanaraps:
Compton doesn't depend on it fyi.

PR #480, opened 2020-02-28
freeglut: new package at 3.2.1

ioraff:
Read: https://k1ss.org/style

---

## Description of package
Provides functionality for small OpenGL programs

## New package


community/freeglut/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -e
+
+export DESTDIR="$1"
+
+cmake -B build \
+    -DCMAKE_INSTALL_PREFIX=/usr \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DCMAKE_BUILD_TYPE=None
```
dylanaraps: Not Release which is probably default?

ioraff: Turns out "None" is the default.

PR #481, opened 2020-02-28
cython: new package at 0.29.15

ioraff:
Read: https://k1ss.org/style

---

## Description of package
C extensions and optimizing compiler for python

## New package


community/cython/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+
+python setup.py build
+python setup.py install \
+    --root="$1" \
+    --skip-build
```
dylanaraps: What does this do?

ioraff: from the Arch Wiki:
> optimizes away the unnecessary attempt to re-run the build steps already run in the build() function

dylanaraps: Nice

PR #497, opened 2020-03-03
libcap: disable GOLANG

a-schaefers:
Read: https://k1ss.org/style

---

## New package

- [ ] I agree to maintain this package.
- [ ] I agree to notify @dylanaraps if I can no longer maintain this package.

## Existing package

### DIFF
--- a/community/libcap/build
+++ b/community/libcap/build
@@ -4,12 +4,16 @@ sed -i "s/CFLAGS :=/CFLAGS := \$(CFLAGS)/;/LDFLAGS :=/d" Make.Rules
 sed -i 's/pthread\.h>/\0 #include <sched.h>/'            libcap/psx.c
 sed -i 's/pthread_yield/sched_yield/'                    libcap/psx.c
 
-make
-make \
-    GOLANG=no \
-    PREFIX=/usr \
-    SBINDIR=/usr/bin \
-    LIBDIR=/usr/lib \
-    DESTDIR="$1" \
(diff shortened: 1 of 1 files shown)


community/libcap/build
```diff
@@ -6,6 +6,7 @@ sed -i 's/pthread_yield/sched_yield/'                    libcap/psx.c
 
 make
 make \
+    GOLANG=no \
     PREFIX=/usr \
     SBINDIR=/usr/bin \
```
dylanaraps: Does this work when run from `make install`?

a-schaefers: wait don't merge this, you caught something here

a-schaefers: should be good now. tested it builds with and without 'go' installed on the system.

community/libcap/build
```diff
@@ -4,11 +4,16 @@ sed -i "s/CFLAGS :=/CFLAGS := \$(CFLAGS)/;/LDFLAGS :=/d" Make.Rules
...
-    DESTDIR="$1" \
-    RAISE_SETFCAP=no \
-    install
+mk() {
+    make \
+        GOLANG=no \
+        PREFIX=/usr \
+        SBINDIR=/usr/bin \
+        LIBDIR=/usr/lib \
+        DESTDIR="$1" \
```
dylanaraps: DESTDIR is duplicated here and below.

a-schaefers: Ah, fixed.

PR #540, opened 2020-03-15
jbig2dec: bump to 0.18

IslandAlpha:
New tarball doesn't have pre-made configure. There is no alternative source that I know of.

## Existing package


community/jbig2dec/build
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -e
 
+./autogen.sh
```
dylanaraps:
```
autoreconf -fi
```

Does this work?

IslandAlpha: Yep, changed

IslandAlpha: yep, fixed

PR #562, opened 2020-03-21
jsoncpp: new package at 1.9.2

IslandAlpha:
## Description of package

JSON library for C++.

The build is a bit unconventional, the CMakeLists.txt won't let me build both static and shared libraries at the same time. Is there even a use having both, or should I build only one of them?

```
/var/db/kiss/installed/jsoncpp/version
/var/db/kiss/installed/jsoncpp/sources
/var/db/kiss/installed/jsoncpp/manifest
/var/db/kiss/installed/jsoncpp/depends
/var/db/kiss/installed/jsoncpp/checksums
/var/db/kiss/installed/jsoncpp/build
/var/db/kiss/installed/jsoncpp/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/lib64/libjsoncpp.so.22
/usr/lib64/libjsoncpp.so.1.9.2
/usr/lib64/libjsoncpp.so
/usr/lib64/libjsoncpp.a
/usr/lib64/
/usr/include/json/writer.h
/usr/include/json/version.h
/usr/include/json/value.h
/usr/include/json/reader.h
/usr/include/json/json_features.h
/usr/include/json/json.h
/usr/include/json/forwards.h
/usr/include/json/config.h
/usr/include/json/autolink.h
/usr/include/json/assertions.h
/usr/include/json/allocator.h
/usr/include/json/
/usr/include/
/usr/
```

## New package


dylanaraps:
What's this needed for? I don't like adding libraries if nothing will immediately use them.

IslandAlpha:
It's actually for a kiss-games package, my reasoning was that since this isn't related to gaming and other community packages might use this library in the future, it's a better fit here than in kiss-games.

IslandAlpha:
Also cmake uses jsoncpp, but the build for it uses a library in its source tree.

IslandAlpha:
I changed my mind, we can just add jsoncpp here in the future.

PR #628, opened 2020-04-02
micro: new package at 2.0.2

onodera-punpun:
##  Description of package

A modern and intuitive terminal-based text editor.

```
/var/db/kiss/installed/micro/version
/var/db/kiss/installed/micro/sources
/var/db/kiss/installed/micro/manifest
/var/db/kiss/installed/micro/depends
/var/db/kiss/installed/micro/checksums
/var/db/kiss/installed/micro/build
/var/db/kiss/installed/micro/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/bin/micro
/usr/bin/
/usr/
```

## New package


community/micro/build
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh -e
+
+export GOPATH="$PWD/go"
+export GO111MODULE=on
+
+# More verbose building.
+sed -i "s/go build/go build -v/" Makefile
```
dylanaraps, line 7: Should we care about this or just leave it default? I'm fine with it either way.

PR #687, opened 2020-04-17
mpd: provide service files. Resolves #646

### LINKED ISSUE #646: mpd: provide service file
Like for example mdev does, in `/ect/sv/`

cemkeylan:
This only adds a service file and an mpd directory to /etc/sv, so that's the only manifest change.


onodera-punpun:
I see many other service files using `exec`, why not this one?

cemkeylan:
That's a mistake, I missed that one.

PR #775, opened 2020-05-03
gstreamer gst-plugins: add missing dependencies

### LINKED ISSUE #774: gstreamer: missing dependency
@cemkeylan 
Gstreamer fails to build without bison.

cemkeylan:
This resolves #774 . No manifest change is made.

Also, I have noticed that gst-plugins depend on linux-headers. So I have
added that as well.

### DIFF
--- a/community/gst-plugins/depends
+++ b/community/gst-plugins/depends
@@ -13,7 +13,7 @@ libsoup
 libvorbis
 libvpx
 libwebp
-linux-headers
+linux-headers make
 meson make
 opus
 pango
--- a/community/gst-plugins/depends
+++ b/community/gst-plugins/depends
@@ -13,6 +13,7 @@ libsoup
 libvorbis
 libvpx
 libwebp
+linux-headers
 meson make
 opus
 pango
--- a/community/gstreamer/depends
+++ b/community/gstreamer/depends
@@ -1,2 +1,3 @@
+bison make
 glib
 meson make


community/gst-plugins/depends
```diff
@@ -13,6 +13,7 @@ libsoup
 libvorbis
 libvpx
 libwebp
+linux-headers
```
dylanaraps, line 16: This is an implicit dependency throughout the repositories.

dylanaraps: ie, it can be removed.

cemkeylan: Isn't bison the same?

konimex: I'd personally say not really, given that a lot of tools can be built without the need of yacc (in KISS' case, bison)

cemkeylan: But a lot of tools can be built with the need of linux-headers

cemkeylan: In fact I have accidentally butchered my build file for linux-headers a few days ago (which caused no headers to be installed) and it went unnoticed until I builded gst-plugins.

konimex: Interesting. Looks like I'll try to build my LLVM stack without linux-headers as an experiment and see if it doesn't need headers to build (one doesn't need yacc there so any dependencies to yacc are explicit instead of implicit).

cemkeylan: Busybox depends on linux-headers if you build it with kernel utilities such as `modprobe`, `ps`. `e2fsprogs` also depend on linux-headers. But the base can be stripped from linux-headers other than that.

dylanaraps: The inclusion of `linux-headers` in the depends file wouldn't have solved your issue regardless no? The same kind of breakage can occur with broken build files for `gcc`, `make` and a myriad of other packages.

This all being said, it'd be nice to figure out exactly which packages require the headers (I'll get onto this).

dylanaraps: Also, should this dependency be `linux-headers` or `linux-headers make`?

cemkeylan: Yeah, I have mentioned my broken package as an example for how it didn't affect most of my packages.

My point was "why should we treat bison and linux headers differently while they are both part of the base". They are both not removed in `kiss-reset` either. Not many packages depend on it as well. That's my personal opinion though.

And you are right, it is a make dependency. Should I remove it or should I append a make?

cemkeylan: I do think that unchangable dependencies on the base (like linux-headers, bison) should be noted as dependencies, while interchangable stuff (like libc, cc, libtls) shouldn't be noted as dependencies.

dylanaraps: Yeah. I think we should move in this direction. See: https://github.com/kisslinux/repo/issues/185

dylanaraps: The one issue is that it's tricky to discern whether or not something depends on the headers without trying a build without them installed.

dylanaraps: (`linux-headers make` to depends in essense)

dylanaraps: @konimex LLVM utilizes `linux-headers` in a single file. It's behind an `ifdef` and includes the headers only if they're found. Clang doesn't require them at all.

dylanaraps:
Thanks

PR #830, opened 2020-05-19
lilo: some fixes from gentoo

onodera-punpun:


community/lilo/build
```diff
@@ -20,7 +20,7 @@ sed -i "/images all/d"                     Makefile src/Makefile
 
 export LC_ALL=C
 
-make CFLAGS="$CFLAGS -fcommon" all
+make CFLAGS="-fcommon" all
```
dylanaraps, line 23: Do we want to clear the user's CFLAGS?

onodera-punpun: Gentoo explicitly removes them for stability reasons.

PR #870, opened 2020-05-30
w3m: bump release (newer source, same version)

trbednarzyk:
As it turns out, there has been a newer source for almost a month but I didn't notice because the version number (0.5.3) didn't change so Repology never said it was outdated. Now that I know that this can happen with w3m, I'll try to pay better attention to keeping it up to date.

## Existing package


dylanaraps:
:+1:

PR #880, opened 2020-05-30
bspwm: fix gcc 10 (Untested on GCC)

konimex:
I can't confirm the current problem since I don't use GCC, but if sxhkd is affected, then bspwm might be affected too. Dylan, can you check if the current packaging without CFLAGS fix work? If it works, then you can close this PR without merging. Thanks.

No changes.

## Existing package


dylanaraps:
Yup. This is needed.

PR #884, opened 2020-05-30
sdl2: build fails with dash fix 

### LINKED ISSUE #872: sdl2: Cannot build with dash
sdl2 cannot build when /usr/bin/sh is set to dash (`kiss a dash /usr/bin/sh`)

This issue is already reported upstream at https://bugzilla.libsdl.org/show_bug.cgi?id=4261, but a workaround (like forcing busybox sh) or warning could be nice.

sdsddsd1:
Should print a warning for #872 

## Existing package

### DIFF
--- a/community/sdl2/build
+++ b/community/sdl2/build
@@ -1,5 +1,20 @@
 #!/bin/sh -e
 
+# Check if dash is installed and print a warning
+kiss l dash &&
+cat <<EOF
+
+### IMPORTANT ###
+### IMPORTANT ###
+
+Note: SDL2 cannot be build with dash. Your build may fail.
+
+### IMPORTANT ###
+### IMPORTANT ###
+
+EOF
+sleep 5
+
 ./configure \
     --prefix=/usr
 
--- a/community/sdl2/build
+++ b/community/sdl2/build
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-# Check if dash is installed and print a warning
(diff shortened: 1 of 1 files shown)


dylanaraps:
Fails in dash but works in busybox sh?

sdsddsd1:
Yep. I build with busybox sh with no problems.

dylanaraps:
I don't like this solution as it affects users with dash installed (but with something else as /bin/sh). Does `CONFIG_SHELL=/bin/sh ./configure \` work at all?

sdsddsd1:
`CONFIG_SHELL= ./configure \` does work.
When busybox is /bin/sh I can make it fail. And when dash is /bin/sh I can make it succed.
We could modify the warning to only print when dash is /bin/sh but I have no nice idea how to query that.

sdsddsd1:
I have found a way to query dash and put that in line with other kiss-tools.
```
#!/bin/sh -e

log() {
    printf '\033[31;1m->\033[m %s.\n' "$@"
}

die() {
    log "$@"
    exit 1
}

kiss l dash &&
kiss a | grep "dash /usr/bin/sh" ||
die "ERROR: sdl2 can not be build with dash. Please change /usr/bin/sh to something appropiate"

./configure \
    --prefix=/usr

make
make DESTDIR="$1" install
```

dylanaraps:
This should be fixed now. Simply update dash.

PR #943, opened 2020-06-15
Fix dependancies for tuitube and curlpp

djt3:
- [Y] I am the maintainer of this package.


community/curlpp/depends
```diff
@@ -1,2 +1,2 @@
-curl
 cmake
```
sdsddsd1, line 1: This should also be `cmake make`

djt3: thanks cant believe i didnt spot that mistake

PR #977, opened 2020-07-02
bitfetch: update to 3.1

jedahan:
Note: the 3.1 archive is missing an update to the makefile which makes it incorrectly report the version number, which is why I pin it to the release commit archive.


dylanaraps:
You aren't the maintainer.

PR #1003, opened 2020-07-08
qt5 no longer requires icu

dilyn-corner:
As per the resolution of [this](https://bugreports.qt.io/browse/QTBUG-85453) bug report, `qt5` only required `icu` because `kiss` set `LC_ALL=C`. As this change has been reverted, we can revert our `icu` requirement.

## Existing package


dylanaraps:
Hooray!

PR #1017, opened 2020-07-10
boost: apply patch for fixing freecad build

cemkeylan:
## Description of package

This is a bug-fix patch that is accepted by upstream, but isn't available as a release
yet.

(`cat /var/db/kiss/installed/<pkg_name>/manifest`)

As usual, the manifest of boost is too big to be added here.
<https://termbin.com/soum>

## Existing package


dylanaraps:
This is far better suited for a simple `sed` call no?

cemkeylan:
Yeah, you are right!

cemkeylan:
I forgot checksums whoops

PR #1043, opened 2020-07-15
github-cli: new package at 0.10.1

mmatongo:
---

## Description of package

GitHub’s official command line tool
## Installed manifest of package

```
/var/db/kiss/installed/github-cli/version
/var/db/kiss/installed/github-cli/sources
/var/db/kiss/installed/github-cli/manifest
/var/db/kiss/installed/github-cli/depends
/var/db/kiss/installed/github-cli/checksums
/var/db/kiss/installed/github-cli/build
/var/db/kiss/installed/github-cli/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/share/zsh/site-functions/_gh
/usr/share/zsh/site-functions/
/usr/share/zsh/
/usr/share/licenses/github-cli
/usr/share/licenses/
/usr/share/bash-completion/completions/gh
/usr/share/bash-completion/completions/
/usr/share/bash-completion/
/usr/share/
/usr/bin/gh
/usr/bin/
/usr/
```

## New package

- [Y] Latest upstream version.
- [Y] I agree to maintain this package.
- [Y] I agree to notify @dylanaraps if I can no longer maintain this package.

PR #1047, opened 2020-07-15
github-cli: new package at 0.10.1

mmatongo:
## Description of package

GitHub’s official command line tool

```
/var/db/kiss/installed/github-cli/version
/var/db/kiss/installed/github-cli/sources
/var/db/kiss/installed/github-cli/manifest
/var/db/kiss/installed/github-cli/depends
/var/db/kiss/installed/github-cli/checksums
/var/db/kiss/installed/github-cli/build
/var/db/kiss/installed/github-cli/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/share/zsh/site-functions/_gh
/usr/share/zsh/site-functions/
/usr/share/zsh/
/usr/share/licenses/github-cli
/usr/share/licenses/
/usr/share/bash-completion/completions/gh
/usr/share/bash-completion/completions/
/usr/share/bash-completion/
/usr/share/
/usr/bin/gh
/usr/bin/
/usr/
```

## New package

- [Y] Latest upstream version.
- [Y] I agree to maintain this package.
- [Y] I agree to notify @dylanaraps if I can no longer maintain this package.


mmatongo:
@dylanaraps to re-echo what I said here #1043 
Yes and No. If you like the convenience of completion then yes if you can live without it then no, I assumed the former as it is something you can negligibly live with.

community/github-cli/sources
```diff
@@ -0,0 +1 @@
+git+https://github.com/cli/cli#15a7ab6b92ec25da2d486fd235b6996857f95772
```
dylanaraps, line 1: Why use git for sources instead of a release tarball?

mmatongo: only real reason is to avoid the `fatal: not a git repository (or any of the parent directories): .git` that pops up at compile time.

dylanaraps: Gotcha.

PR #1067, opened 2020-07-21
pyqt5: new package at 5.15.0

sdsddsd1:
## Description of package
PyQt5 is a comprehensive set of Python bindings for Qt v5.

```
https://termbin.com/zgel4
```

## New package


community/pyqt5/build
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh -e
+
+python configure.py \
+    --confirm-license
+
+# Build single threaded to avoid racecondition
```
dylanaraps, line 6: This only *installs* as single threaded no? The build still uses `MAKEFLAGS` right?

sdsddsd1: Oh yes. The buildfile instructions are fine. The comment so lala.

PR #1080, opened 2020-07-24
musl-fts: new package at 1.2.7

dm9pZCAq:
## Description of package
The **musl-fts** package implements the `fts(3)` functions `fts_open`, `fts_read`, `fts_children`, `fts_set` and `fts_close`, which are missing in `musl libc`.

```
/var/db/kiss/installed/musl-fts/version
/var/db/kiss/installed/musl-fts/sources
/var/db/kiss/installed/musl-fts/manifest
/var/db/kiss/installed/musl-fts/depends
/var/db/kiss/installed/musl-fts/checksums
/var/db/kiss/installed/musl-fts/build
/var/db/kiss/installed/musl-fts/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/share/man/man3/musl-fts.3
/usr/share/man/man3/
/usr/share/man/
/usr/share/
/usr/lib/pkgconfig/musl-fts.pc
/usr/lib/pkgconfig/
/usr/lib/libfts.so.0.0.0
/usr/lib/libfts.so.0
/usr/lib/libfts.so
/usr/lib/libfts.la
/usr/lib/libfts.a
/usr/lib/
/usr/include/fts.h
/usr/include/
/usr/
```

## New package


dylanaraps:
I'm against adding ripped out code from other C libraries to KISS.

Also see: https://wiki.musl-libc.org/faq.html#Q:-Why-is-%3Ccode%3Efts.h%3C/code%3E-not-included?

> It’s impossible to match the glibc ABI for fts because the glibc ABI is unusably broken. This also means that no working software can be built using the built-in fts from glibc; anything using fts is either already including its own copy (there is a canonical BSD version and gnulib has it too), or is hopelessly broken on glibc and the maintainers are just unaware of that fact. So including it in musl would not help anything at this time. If glibc bug 15838 is fixed by adding an fts64 interface in glibc, we could consider supporting it with a matching ABI in musl, but it seems more likely that glibc will just deprecate this interface.

Thanks for opening these pull requests. :+1:

Issue #1150, opened 2020-08-24
libcap build failure


periish:
libcap fails to build, with an error message of `indent: Command line: unknown parameter "-kr"`.
@a-schaefers

a-schaefers:
Can anyone else confirm this is true? I thought I test stuff.

dylanaraps:
I can reproduce. 

```sh
-> libcap Starting build
make -C libcap all
make[1]: Entering directory '/tmp/build-1491/libcap/libcap'
=> making cap_names.list.h from /tmp/build-1491/libcap/libcap/../libcap/include/uapi/linux/capability.h
perl -e 'while ($l=<>) { if ($l =~ /^\#define[ \t](CAP[_A-Z]+)[ \t]+([0-9]+)\s+$/) { $tok=$1; $val=$2; $tok =~ tr/A-Z/a-z/; print "{\"$tok\",$val},\n"; } }' /tmp/build-1491/libcap/libcap/../libcap/include/uapi/linux/capability.h | fgrep -v 0x > cap_names.list.h
sed -e 's,@prefix@,/usr,' \
        -e 's,@exec_prefix@,,' \
        -e 's,@libdir@,/usr/lib,' \
        -e 's,@includedir@,/usr/include,' \
        -e 's,@VERSION@,2.42,' \
        -e 's,@deps@,,' \
        libcap.pc.in >libcap.pc
sed -e 's,@prefix@,/usr,' \
        -e 's,@exec_prefix@,,' \
        -e 's,@libdir@,/usr/lib,' \
        -e 's,@includedir@,/usr/include,' \
        -e 's,@VERSION@,2.42,' \
        -e 's,@deps@,,' \
        libpsx.pc.in >libpsx.pc
cc -O2 -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -fPIC -I/tmp/build-1491/libcap/libcap/../libcap/include/uapi -I/tmp/build-1491/libcap/libcap/../libcap/include _makenames.c -o _makenames
[...]

PR #1187, opened 2020-09-01
libcap: bump to 2.43 & fix for #1150

mmatongo:
# Existing package

- [Y] I am the maintainer of this package.


community/libcap/build
```diff
@@ -11,5 +11,5 @@ mk() {
         "$@"
 }
 
-mk
-mk DESTDIR="$1" install
+make INDENT=true
+make DESTDIR="$1" install
```
dylanaraps, line 15: You can drop the `mk` function in this case.

dylanaraps: I'd like to see the resulting manifest too. :+1:

mmatongo: done, new pr opened.

PR #1193, opened 2020-09-02
libgpg-error: 1.39

onodera-punpun:


community/libgpg-error/version
```diff
@@ -1 +1 @@
-1.38 1 
+1.39 1
```
dylanaraps, line 1: Missing newline here?

PR #1207, opened 2020-09-06
iproute2: enforce --prefix/usr and fix build on clang only systems

magenbluten:
`iproute2`'s makefile uses hardcoded `gcc` in one place even if it should use at least `$CC`. if gcc is installed you get a frankenbuild with mixed `$CC` ( clang ) and `gcc`. This change fixes this.

## Existing package

### DIFF
--- a/community/iproute2/build
+++ b/community/iproute2/build
@@ -2,7 +2,13 @@
 
 patch -p1 < 0001-make-iproute2-fhs-compliant.patch
 
-./configure
+# Fix hardcoded `gcc` ( there is HOST_CC := $(CC) on line after this,
+# but upstream seems to use it inconsistently ). Otherwise build fails on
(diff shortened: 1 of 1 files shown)


community/iproute2/build
```diff
@@ -2,7 +2,13 @@
 
 patch -p1 < 0001-make-iproute2-fhs-compliant.patch
 
-./configure
+# Fix hardcoded `gcc` ( there is HOST_CC := $(CC) on line after this,
+# but upstream seems to use it inconsistently ). Otherwise build fails on
+# `clang` based systems with no `gcc` installed.
```
dylanaraps, line 7: Add `# shellcheck disable=2016` to silence the CI errors.

dylanaraps:
Didn't notice the force push. :)

PR #1236, opened 2020-09-12
Elvish: new package at 0.14.1

monarrk:
Elvish combines usable features of regular shells and syntax and features which make programming in it feel like using any other scripting language. It also has other useful programming features, such as a module system, command line syntax highlighting, and a powerful editing library.

```
/var/db/kiss/installed/elvish/version
/var/db/kiss/installed/elvish/sources
/var/db/kiss/installed/elvish/manifest
/var/db/kiss/installed/elvish/depends
/var/db/kiss/installed/elvish/checksums
/var/db/kiss/installed/elvish/build
/var/db/kiss/installed/elvish/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
```

## New package

- [ * ] Latest upstream version.
- [ * ] I agree to maintain this package.
- [ * ] I agree to notify @dylanaraps if I can no longer maintain this package.


dylanaraps:
The manifest is incorrect. No files are installed at all?

community/elvish/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh -e
+
+:> nostrip
+
+make
```
dylanaraps, line 5: You must also run `make install` with `DESTDIR` set to `$1`.

monarrk:
because it's written in go, it installs to `$HOME/go/bin` or `$GOPATH` which I don't believe is picked up by the manifest. I was willing to leave it that way but if there's a way to make it install to /usr/bin I'd of course be willing to

dylanaraps:
Yes. 

```sh
install -Dm755 elvish/elvish "$1/usr/bin/elvish"
```

Something like this is required if the Makefile does not support it.

dylanaraps:
If the Makefile puts it in `$HOME`, don't run the Makefile.

dylanaraps:
Note: The package manager is useless if packages go installing things in `$HOME` during builds (it is bypassed entirely).

monarrk:
Ok, I just updated the build to use `go build` directly rather than make. here's the new manifest:
```/var/db/kiss/installed/elvish/version
/var/db/kiss/installed/elvish/sources
/var/db/kiss/installed/elvish/manifest
/var/db/kiss/installed/elvish/depends
/var/db/kiss/installed/elvish/checksums
/var/db/kiss/installed/elvish/build
/var/db/kiss/installed/elvish/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/bin/elvish
/usr/bin/
/usr/```

dylanaraps:
Nice :+1:

PR #1245, opened 2020-09-14
cfm: bump to 0.6.4

willeccles:
## Description of package
cfm is a simple and fast TUI file manager with no dependencies. This update adds page up and page down keys as well as a slight change in drawing/updating the screen.

## Existing package


dylanaraps:
For future reference, checksums files should have two spaces between the elements rather than one. Just a stylistic thing.

willeccles:
> 
> 
> For future reference, checksums files should have two spaces between the elements rather than one. Just a stylistic thing.

Interesting. I assume this is just for ease of reading?

dylanaraps:
Older versions of the package manager required `<checksum>  <basename of file>` (two spaces). Newer versions of the package manager only require `<checksum>`. The second field was quite useless as the file contents are ordered anyway. The package manager supports both formats now (but will generate the new format via `kiss c`) by simply comparing only the first field in each file.

dylanaraps:
Ideally you'd drop the second field in checksums entirely. Running `kiss c` will do all of this for you.

willeccles:
Ah, I thought `kiss c` (which is what I got the checksum from) was just generating the checksum, to be added to the file manually later. This is good to know. In the next version I'll remove the second field.

dylanaraps:
`kiss c` will also write the checksums to the file for you, yeah. It handles everything in other words.

willeccles:
It didn't write the file for me, which why I thought it was meant to be copied.

dylanaraps:
What version of the package manager are you using? I can't reproduce this issue and I've also received countless PRs since the changes were made.

willeccles:
I just installed it this morning, so whatever was in master at about 10:30 (-0400).

willeccles:
It claims it's 5.1.0 (sorry for the delay, was doing something else).

willeccles:
If it helps, my kiss root is ~/kissroot and I have my repo cloned to ~/kissroot/repos/community, and did the updates to ~/kissroot/repos/community/community/cfm/*.

I also was not in the cfm directory when I ran the command, so if that makes a difference that could be it.

PR #1257, opened 2020-09-16
secret: new package at git

angt:
Signed-off-by: Adrien Gallouët <adrien@gallouet.fr>

Please fill out the template.

Read: https://k1ss.org/wiki/kiss/style-guide

---

## Description of package

A simple and tiny tool that will help you keep your little secrets.

## Features

`secret` is the simplest secret store you can think of.
But it does have some interesting features:

 - Requires only one file `~/.secret` that you can share publicly without fear.
 - No configuration. No directory. Get back your file and you're done.
 - Secret names (usually hostname, mail, login, etc.) are also encrypted.
 - A secret agent that only trusts subprocesses. Not all the processes of the same user!
 - Secret names completion is available after calling the secret agent.
 - Supports unstored secrets. Derived from some simple keys and a passphrase.
 - Supports multiple passphrases. A confirmation is requested for each new passphrase.
 - Depends only on the [libhydrogen](https://libhydrogen.org/) library.
 - Small, simple and non obfuscated C code. Well, I hope so :)

(`kiss manifest <pkg>`)

```
/var/db/kiss/installed/secret/version
/var/db/kiss/installed/secret/sources
/var/db/kiss/installed/secret/manifest
/var/db/kiss/installed/secret/checksums
/var/db/kiss/installed/secret/build
/var/db/kiss/installed/secret/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/local/bin/secret
/usr/local/bin/
/usr/local/
/usr/
```

## New package

## Existing package


dylanaraps:
Binaries should be in `/usr/bin` and not `/usr/local/bin`

community/secret/build
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
+
+make DESTDIR="$1" install
```
dylanaraps, line 3:
```sh
make
make DESTDIR="$1" PREFIX=/usr install
```

This should also fix the install location issue.

community/secret/checksums
```diff
@@ -0,0 +1 @@
+git  git+https://github.com/angt/secret
```
dylanaraps, line 1: Checksums file is not required when there are only git sources.

dylanaraps:
Build fails on this machine: 
```sh
cc -Wall -O2   secret.c -o secret
secret.c:18:10: fatal error: argz/argz.c: No such file or directory
   18 | #include "argz/argz.c"
      |          ^~~~~~~~~~~~~
compilation terminated.
```

Build file:

```sh
#!/bin/sh -e

make
make DESTDIR="$1" prefix=/usr install
```

dylanaraps:
I see that both dependencies are vendored but the Makefile provides no means of automatically making use of these if not available in the system Further, the package manager doesn't automatically pull down submodules so the build file will need to do this.

angt:
Oh.. I saw this https://github.com/kisslinux/kiss/pull/172 thinking it was merged ?

dylanaraps:
Actually, this is all that is needed:

```sh
#!/bin/sh -e

git submodule update --remote --init

make
make DESTDIR="$1" prefix=/usr install
```

konimex:
The PR you mentioned is for the *repo*, not for the packages.

angt:
Yes, that makes sense. Doing it in the build script is a better solution 👍

dylanaraps:
That PR was reverted later as it added immense clone times for monorepos where submodules (which weren't needed) were pulled down in addition to the repository. It's much better as an opt-in thing (git repositories aren't checksummed anyway). 

There is another option both for git based sources and tarballs.

```sh
# All git sources.
git+https://github.com/angt/secret
git+https://github.com/angt/argz#ea7d185d496feecd052f16e141318dfcc6156c69 argz
git+https://github.com/jedisct1/libhydrogen#c879a1d572de4bb2d3bd90516f449557eb156c96 libhydrogen

# Tarballs for vendored sources (can use tarball for secret too if desired).
git+https://github.com/angt/secret
https://github.com/angt/argz/archive/ea7d185d496feecd052f16e141318dfcc6156c69.tar.gz argz
https://github.com/jedisct1/libhydrogen/archive/c879a1d572de4bb2d3bd90516f449557eb156c96.tar.gz libhydrogen
```

dylanaraps:
LGTM

dylanaraps:
One more thing; any reason why this isn't following tagged releases?

angt:
I removed the checksums file now `kiss` is unhappy:
```
-> secret Checking repository files
ERROR secret Checksums are missing
```
Did I misunderstood something ?

dylanaraps:
@angt What version of the package manager do you have?

angt:
`kiss version` shows 5.0.6

angt:
It's a really small project, I think `master` will always be stable from now on.

dylanaraps:
Latest version is `5.1.0` (`5.0.6` is one behind). A blank checksums file is a fix for this (at least as a temporary measure until everyone is running `5.1.0`). This is the first package to not contain checksums in the KISS repositories so I'm glad we've caught this now. 

> It's a really small project, I think master will always be stable from now on.

I was more concerned about this being a security focused tool where tagged releases and checksum verification become a more desirable thing. I'm OK with this either way, I'll leave the choice to you.

angt:
You right, but I couldn't find a way to git clone a specific tag (@ and # only works for branches).

dylanaraps:
I meant: https://github.com/angt/secret/archive/v0.10.tar.gz

dylanaraps:
Sources file would be: 

```sh
https://github.com/angt/secret/archive/v0.10.tar.gz
https://github.com/angt/argz/archive/ea7d185d496feecd052f16e141318dfcc6156c69.tar.gz argz
https://github.com/jedisct1/libhydrogen/archive/c879a1d572de4bb2d3bd90516f449557eb156c96.tar.gz libhydrogen
```

All sources will be checksummed this way also.

PR #1228, opened 2020-09-10
usbutils: add libtool make dependency

jedahan:
I needed this dep to build correctly.
```
/var/db/kiss/installed/usbutils/version
/var/db/kiss/installed/usbutils/sources
/var/db/kiss/installed/usbutils/manifest
/var/db/kiss/installed/usbutils/depends
/var/db/kiss/installed/usbutils/checksums
/var/db/kiss/installed/usbutils/build
/var/db/kiss/installed/usbutils/
/var/db/kiss/installed/
/var/db/kiss/choices/
/var/db/kiss/
/var/db/
/var/
/usr/share/man/man8/usbhid-dump.8
/usr/share/man/man8/lsusb.8
/usr/share/man/man8/
/usr/share/man/man1/usb-devices.1
/usr/share/man/man1/
/usr/share/man/
/usr/share/
/usr/bin/usbhid-dump
/usr/bin/usb-devices
/usr/bin/lsusb.py
/usr/bin/lsusb
/usr/bin/
/usr/
```

## Existing package

- [ ] I am the maintainer of this package.

PR #1281, opened 2020-09-19
usbutils: add libtool make dependency

trbednarzyk:
Just saw #1228. No issue was opened so I didn't notice anything until now.

## Existing package


dylanaraps:
Just for next time, no need to bump the second field for a simple change to the depends file. :+1:

PR #1288, opened 2020-09-22
pciutils: manually handle libraries

### LINKED ISSUE #1283: pciutils: libpci.so.3 links to libpci.a instead of libpci.so.3.7.0
On my system, libpci.so.3 symlinks to libpci.a instead of libpci.so.3.7.0.

Can you reproduce? @cemkeylan

cemkeylan:
This should finally fix all the issues with the shared libraries, we are dealing with the .so creation manually. Closes #1283


dylanaraps:
You can also just do this: 
```sh
#!/bin/sh -e

mk() {
    make \
        CC="${CC:-cc}" \
        OPT="$CFLAGS -fPIC" \
        ZLIB=yes \
        PREFIX=/usr \
        SBINDIR=/usr/bin \
        SHAREDIR=/usr/share/hwdata \
        "$@"
}

mk
install -Dm644 lib/libpci.a "$1/usr/lib/libpci.a"

make clean

mk SHARED=yes
mk SHARED=yes DESTDIR="$1" install
```

cemkeylan:
That didn't work for shared libraries, the Makefile still linked objects to the static library.

dylanaraps:
What about doing it like this?

```sh
#!/bin/sh -e

mk() {
    make \
        CC="${CC:-cc}" \
        OPT="$CFLAGS -fPIC" \
        ZLIB=yes \
        PREFIX=/usr \
        SBINDIR=/usr/bin \
        SHAREDIR=/usr/share/hwdata \
        "$@"
}


for shared in yes no; do (
    cd "$shared"

    mk SHARED="$shared"
    mk SHARED="$shared" DESTDIR="$1" install install-pcilib
) done
```

Sources:

```
https://www.kernel.org/pub/software/utils/pciutils/pciutils-3.7.0.tar.gz yes
https://www.kernel.org/pub/software/utils/pciutils/pciutils-3.7.0.tar.gz no
```

(Source will only be downloaded once)

dylanaraps:
I should be clearer, I don't like this solution. There should be a better way. Will look into it further and hopefully come up with something. :+1:

PR #1297, opened 2020-09-23
glib-networking: bump to 2.66.0

cemkeylan:
## Changes

- The `openssl` implementation on glib-networking has been rewritten. It now
  depends on the OpenSSL TLS1.3 API, which is not fully implemented on libressl
  yet. This leaves us only the choice to use GnuTLS for building glib-networking.
  The dependency file is updated accordingly.

(`kiss manifest <pkg>`)
```
/var/db/kiss/installed/glib-networking/version
/var/db/kiss/installed/glib-networking/sources
/var/db/kiss/installed/glib-networking/manifest
/var/db/kiss/installed/glib-networking/depends
/var/db/kiss/installed/glib-networking/checksums
/var/db/kiss/installed/glib-networking/build
/var/db/kiss/installed/glib-networking/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/lib/gio/modules/libgiognutls.so
/usr/lib/gio/modules/
/usr/lib/gio/
/usr/lib/
/usr/
```

## Existing package


dylanaraps:
Will merge but we should swap back to libressl when possible (client/server side tls 1.3 is done, only thing missing is openssl API for it).

PR #1300, opened 2020-09-24
iana-etc: new package at 20200910

konimex:
## Description of package

Unix /etc/protocols and /etc/services files. Needed for OpenNTPD otherwise the daemon will fail.

It'll always show up as "outdated" on repology as there's no single source for it.

```
/var/db/kiss/installed/iana-etc/version
/var/db/kiss/installed/iana-etc/sources
/var/db/kiss/installed/iana-etc/manifest
/var/db/kiss/installed/iana-etc/files/services
/var/db/kiss/installed/iana-etc/files/protocols
/var/db/kiss/installed/iana-etc/files/
/var/db/kiss/installed/iana-etc/etcsums
/var/db/kiss/installed/iana-etc/checksums
/var/db/kiss/installed/iana-etc/build
/var/db/kiss/installed/iana-etc/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/etc/services
/etc/protocols
/etc/
```

## New package

- [ ] Latest upstream version. (No idea)


dylanaraps:
It's a shame that we must package it like this but anyway.

PR #1302, opened 2020-09-24
yarn: new package at 1.22.5

mmatongo:
## Description of package
Fast, reliable, and secure dependency management for Node.js

```
/var/db/kiss/installed/yarn/version
/var/db/kiss/installed/yarn/sources
/var/db/kiss/installed/yarn/manifest
/var/db/kiss/installed/yarn/depends
/var/db/kiss/installed/yarn/checksums
/var/db/kiss/installed/yarn/build
/var/db/kiss/installed/yarn/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/share/node_modules/yarn/package.json
/usr/share/node_modules/yarn/lib/v8-compile-cache.js
/usr/share/node_modules/yarn/lib/cli.js
/usr/share/node_modules/yarn/lib/
/usr/share/node_modules/yarn/bin/yarnpkg
/usr/share/node_modules/yarn/bin/yarn.js
/usr/share/node_modules/yarn/bin/yarn
/usr/share/node_modules/yarn/bin/
/usr/share/node_modules/yarn/
/usr/share/node_modules/
/usr/share/
/usr/bin/yarnpkg
/usr/bin/yarn
/usr/bin/
/usr/
```

## New package

- [Y] Latest upstream version.
- [Y] I agree to maintain this package.
- [Y] I agree to notify @dylanaraps if I can no longer maintain this package.


mmatongo:
Worth a tr

mmatongo:
Don't judge me :sweat_smile:

community/yarn/build
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh -e
+
+mkdir -p "$1/usr/bin"
+mkdir -p "$1/usr/share/node_modules/yarn"
+
+rm -rf bin/*.cmd
+rm LICENSE
+rm README.md
+
+cp -a ./* "$1/usr/share/node_modules/yarn"
```
dylanaraps, line 10: `-a` isn't POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html

mmatongo: fixed

community/yarn/build
```diff
@@ -7,7 +7,7 @@ rm -rf bin/*.cmd
 rm LICENSE
 rm README.md
 
-cp -a ./* "$1/usr/share/node_modules/yarn"
+cp -r ./* "$1/usr/share/node_modules/yarn"
```
dylanaraps, line 10: Still not POSIX: 

```
Earlier versions of this standard included support for the -r option to copy file hierarchies. The -r option is historical practice on BSD and BSD-derived systems. This option is no longer specified by POSIX.1-2017 but may be present in some implementations. The -R option was added as a close synonym to the -r option, selected for consistency with all other options in this volume of POSIX.1-2017 that do recursive directory descent.

The difference between -R and the removed -r option is in the treatment by cp of file types other than regular and directory. It was implementation-defined how the - option treated special files to allow both historical implementations and those that chose to support -r with the same abilities as -R defined by this volume of POSIX.1-2017. The original -r flag, for historic reasons, did not handle special files any differently from regular files, but always read the file and copied its contents. This had obvious problems in the presence of special file types; for example, character devices, FIFOs, and sockets.
```

mmatongo: Aaaah, I see.

PR #1326, opened 2020-09-30
osh: bump to 0.8.1

jedahan:
## Existing package


community/osh/build
```diff
@@ -3,7 +3,7 @@
 :> nostrip
 
 ./configure \
-    --prefix=/usr
+    --prefix /usr
```
dylanaraps, line 6: Intentional?

dylanaraps: Usage shows `=` https://github.com/oilshell/oil/blob/master/configure#L38

jedahan: I was following the documentation http://www.oilshell.org/release/latest/doc/INSTALL.html I will ask the maintainers of osh.

dylanaraps: Both work. I think it's best to use `=` anyway. I don't know why both syntaxes are even supported... the usage in this PR is undocumented too.

dylanaraps: See: https://github.com/oilshell/oil/blob/master/configure#L96-L107

PR #1330, opened 2020-10-01
bat: new package at 0.15.4

git-bruh:
## Description of package
A `cat` clone with useful features like syntax highlighting and `git` integration.

(`kiss manifest <pkg>`)

```
/var/db/kiss/installed/bat/version
/var/db/kiss/installed/bat/sources
/var/db/kiss/installed/bat/post-install
/var/db/kiss/installed/bat/manifest
/var/db/kiss/installed/bat/depends
/var/db/kiss/installed/bat/checksums
/var/db/kiss/installed/bat/build
/var/db/kiss/installed/bat/
/var/db/kiss/installed/
/var/db/kiss/
/var/db/
/var/
/usr/share/man/man1/bat.1
/usr/share/man/man1/
/usr/share/man/
/usr/share/
/usr/bin/bat
/usr/bin/
/usr/
```

## New package


community/bat/build
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+cargo build --release
+
+install -Dm755 target/release/bat "$1/usr/bin/bat"
+find . -name bat.1 -exec install -Dm644 {} \
```
dylanaraps, line 6: Do we not know where the file is?

git-bruh: The man-page is generated at a path that looks like this:
`./target/release/build/bat-484c9d3dc1d6eff0/out/assets/manual/bat.1`
If i compile the git version then the path is like this:
`./target/release/build/bat-ef405f4526930eea/out/assets/manual/bat.1`
Hence, `find` is used as the man-page path will probably change over new releases

dylanaraps:
```sh
install -Dm644 target/release/build/bat-*/out/assets/manual/bat.1 "$1/usr/share/man/man1"
```

Does this work?

dylanaraps: Can we not install everything using `cargo`?

git-bruh: Cargo doesn't seem to install the man-page, only the executable is installed.
I've added a commit to use the method you suggested for installing the man-page.

git-bruh: Seems like cargo doesn't yet support installing man-pages either, so it'll have to be done manually. See https://github.com/rust-lang/cargo/issues/2729

Issue #1311, opened 2020-09-27
pkgconf is no longer an implicit dependency. some pkg's don't build without it.


eudaldgr:
`kmod` - @magenbluten 
~~`qemu` - @jedavies-dev~~
~~`libtorrent` - @jedavies-dev~~
~~`transmission` - @onodera-punpun~~ 
~~`girara` - @Vouivre~~
~~`zathura` - @Vouivre~~
~~`gnutls` - @cemkeylan~~
~~`glib-networking` - @cemkeylan~~
~~`libsoup` - @cemkeylan~~
~~`poppler` - @Vouivre~~
`xcb-util-xrm` - @alaughlin 
~~`alsa-plugins` - @onodera-punpun~~
~~`bind` - @aicsx~~
`conky` - @dzove855 
~~`cups-filters` - @Vouivre~~
~~`picom` - @konimex~~
~~`w3m` - @trbednarzyk~~

Vouivre:
I will update my packages in the following days. I began with `cups-filters`, but it doesn't build anymore. I will first fix it.

Vouivre:
@eudaldgr How did you get this list of packages for which `pkgconf` is missing ? 

For the future, if I create a new package, it's possible that I miss this dependency because it's installed by another package. I only see two solutions:

- check the build dependencies in the documentation of the software
- uninstall pkgconf and test the build

But to check for a lot of packages, I expect another method.

eudaldgr:
I have a chroot with all the main repo pkg's installed, since a lot of time for diferent purposes.
On my machine i do a remove orphans regularly, and `pkgconf` is only needed at build time, so when i see i can't install `kmod` and later update `qemu`, i take that chroot, removed `pkgconf` package and tried to install all community pkg's.

That list is not completed i do a part of the community repo only.

So, yes 'uninstall pkgconf and test the build'

Vouivre:
Thank you for your answer! All my packages are updated now and I will check this dependency in the future.

dylanaraps:
Made the remaining changes. Thanks everyone.

eudaldgr:
The list are not exhaustive, maybe there is more out there, but thanks!

dylanaraps:
Yeah... There are two options here:

1. `kiss r pkgconf` -> try to build everything in Community.
2. Wait until bug reports are opened by users.

PR #1397, opened 2020-10-14
w3m: add pkgconf as make dependency

trbednarzyk:
## Existing package

Added `pkgconf` as a make dependency per issue #1311.

On a side note, I noticed on Repology the version changed to 0.5.3pl38 (from 0.5.3). I looked at the source (https://github.com/tats/w3m) and on Repology and believe this to be the same release currently used. It looks like only OpenBSD is using this version name currently.
Should I bump the version so that it shows up as up-to-date on Repology or leave it as-is?

### DIFF
--- a/community/w3m/depends
+++ b/community/w3m/depends
@@ -1,1 +1,2 @@
 gc
+pkgconf make


dylanaraps:
OpenBSD are actually using Debian as the upstream to w3m. Tarball is here: https://github.com/tats/w3m/archive/debian/0.5.3-37.tar.gz Version in this repository should be kept as-is with an issue opened on Repology's tracker about it.

dylanaraps:
w3m has also seen no release since 2011 according to this: https://sourceforge.net/projects/w3m/files/w3m/w3m-0.5.3/ It might be worth swapping to Debian for upstream (if it's worth it of course).

dylanaraps:
Changelog for Debian's w3m: https://github.com/tats/w3m/blob/master/ChangeLog

trbednarzyk:
The tarball you linked is actually to an outdated version (released January 2019 as opposed to the current one released May 2020). We already use the "same" upstream as Debian, https://github.com/tats/w3m/releases, but not the debian branch/release. The release we are using was made on the same day using the master branch, however. Looking the debian branch, it currently claims `307 commits ahead, 1989 commits behind master`. I'm not currently sure what the difference between the two branches are yet but it looks the master branch gets updated more frequently (committed to 16 days ago vs May 25th).

I'll look into it some more and see which release we should be tracking.

PR #1428, opened 2020-10-21
zathura-pdf-mupdf: add gumbo-parser as dep

noocsharp:
## Existing package


dylanaraps:
I'm confused about this change. `gumbo` does not appear anywhere in the repository?

https://github.com/pwmt/zathura-pdf-mupdf/search?q=gumbo

noocsharp:
This is really fixing an issue that's been present after upgrading libmupdf to 1.18 and adding gumbo as a dependency. The patch makes zathura-pdf-mupdf link against gumbo, along with the other already present libmupdf dependencies. If we don't explicitly tell it to link against gumbo, it will build, but it will complain about missing symbols related to gumbo when you try to open a pdf.

PR #1443, opened 2020-10-26
youtube-dl: update checksum

Vouivre:
## Existing package

- [x] I am the maintainer of this package.

PR #1444, opened 2020-10-26
youtube-dl: update checksum (bis)

Vouivre:
## Existing package

If there is the same problem as PR https://github.com/kisslinux/community/pull/1443, just tell me what is wrong. I have 
installed this version and I didn't see which file were installed on `/`.

Thanks!


dylanaraps:
See how it creates `community/version` instead of modifying `community/community/youtube-dl/version`? https://github.com/Vouivre/community/blob/c81530e634764cf6f7d7d81fa5ce1e0ad25646c3/version

onodera-punpun:
I'm still having a checksum mismatch...

dylanaraps:
Looks like we'll need to find a mirror.

Vouivre:
Here perhaps:

https://yt-dl.org/download.html

Do I update this package with this sources ? 

Perhaps we could also drop it to use an alternative. I will probably open a subject on the KISS Linux community on Reddit.

PR #1456, opened 2020-10-29
youtube-dl: update sources

Vouivre:
## Existing package

Please have a look at the repo before merging, it seems to me to be a reliable fork of `youtube-dl`. 
I hope it fixes the checksum problem.


dylanaraps:
We should hold off on updating (or changing which remote we use) until it's clear as to what will happen to youtube-dl. This PR is fine as the checksums are identical for the sources.

Vouivre:
At the beginning I only wanted to update the sources. But there was some problems, the last was the reported checksum mismatch. If it's ok now, I don't want to update it anymore until a solution is found. Thank you for merging!

Issue #1460, opened 2020-10-30
Dropping some packages


eudaldgr:
* [x] ~~2bwm~~ @Vouivre 
* [x] ~~aerc~~ @muevoid 
* [x] ~~birch~~ - @dylanaraps 
* [x] ~~btpd~~ - @noocsharp 
* [x] ~~dejavu-ttf~~ @Vouivre 
* [x] ~~font-awesome-otf~~
* [x] ~~font-awesome-ttf~~
* [x] ~~hack~~ @dylanaraps 
* [x] ~~ledger~~ dropped from community
* [x] ~~libburn~~ @dylanaraps 
* [x] ~~libisoburn~~
* [x] ~~libisofs~~ @dylanaraps 
* [x] ~~radare2~~ dropped from community
* [x] ~~sc~~ @Xeno-Hypster 
* [x] ~~sc-im~~ @Vouivre 
* [x] ~~scdoc~~ @dylanaraps 
* [x] ~~shfm~~ @dylanaraps 
* [x] ~~squashfs-tools~~ @git-bruh 
* [x] ~~surf~~ @mcpcpc 
* [x] ~~svkbd~~ - @nikotidar 
* [x] ~~syslinux~~ dropped from community
* [x] ~~zzz~~ @dylanaraps ~~@onodera-punpun~~ ??

dylanaraps:
I'll take birch and shfm (for obvious reasons) ;)

mcpcpc:
I'll take surf

onodera-punpun:
I'll take `zzz`

Xeno-Hypster:
I'll be happy to take sc.

noocsharp:
I can take btpd

TAAPArthur:
I can take ledger and svkbd

Vouivre:
I will adopt 2bm, sc-im and dejavu-ttf. Unfortunately there are some errors when I build sc-im and dejavu-ttf. I will first fix them.

dylanaraps:
Build fix for dejavu-ttf:

```sh
#!/bin/sh -e

for file in ttf/*.ttf; do
    install -Dm644 "$file" "$1/usr/share/fonts/TTF/${file##*/}"
done
```

dylanaraps:
Build fix for `sc-im`:

```sh
#!/bin/sh -e

export CFLAGS="$CFLAGS -fcommon"

make -C src
make -C src \
    DESTDIR="$1" \
    prefix=/usr \
    install
```

[...]

PR #1519, opened 2020-11-06
sc-im: -fcommon flag added to CFLAGS

Vouivre:
Package adopted (see https://github.com/kisslinux/community/issues/1460). 

`build` file modified by addind the flag `-fcommon` in `CFLAGS`.


dylanaraps:
You still must modify the version file. :+1:

Vouivre:
Ok, done! I understood that a change had to be done to adopt a package. When the package is fine, the file `version` can be changed. Why is it necessary to always change it to adpot a package ?

dylanaraps:
Maintainer is based on latest commit to the version file.

Vouivre:
Thank you!

PR #1577, opened 2020-11-19
libedit: patch for gcc-less systems

trbednarzyk:
## Existing package

### DIFF
--- a/community/libedit/build
+++ b/community/libedit/build
@@ -1,5 +1,10 @@
 #!/bin/sh -e
 
+# Allows compilation as long as wchar_t is 32 bits (signed or unsigned), even if
+# __STDC_ISO_10646__ isn't defined, which seems to be the case for musl + LLVM
+# without gcc.
+sed -i 's/#ifndef __STDC_ISO_10646__/#if WCHAR_MAX < 2147483647/' src/chartype.h
+
 ./configure \
     --prefix=/usr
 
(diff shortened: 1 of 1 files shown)


community/libedit/build
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh -e
 
+# Allows compilation as long as wchar_t is 32 bits (signed or unsigned), even if
+# __STDC_ISO_10646__ isn't defined, which seems to be the case for musl + LLVM
+# without gcc.
+sed -i 's/#ifndef __STDC_ISO_10646__/#if WCHAR_MAX < 2147483647/' src/chartype.h
+
```
dylanaraps, line 7: Musl defines this.  Just add this header to `src/chartype.h`: `#include <stdc-predef.h>`.

trbednarzyk: You're right. I had actually checked this file earlier but I somehow missed that particular line. I'll fix that in a second.

dylanaraps: However, it seems that llvm/clang are supposed to automatically include this file if it exists: https://reviews.llvm.org/D34158

dylanaraps: GCC also pre-includes it (which is why GCC builds work). See: https://gcc.gnu.org/gcc-4.8/porting_to.html

dylanaraps: Yup. These made it upstream:: https://github.com/llvm/llvm-project/commit/5c086c7626fab2617c0178a3951378bcbce2271c https://github.com/llvm/llvm-project/commit/bb322555afb4fb135abc48fbb3383b7216557995

trbednarzyk: Strange that it's still happening then. I'll look into why it's not getting included automatically later (probably next week).

